PR #101, opened 2022-03-30
Exclude scaffolding templates when package app

rainboyan:
If install scaffold templates, when package app war the templates also include `build/resources/main/META-INF/templates`. These templates should not be included.


puneetbehl:
I think that it is correct behaviour because one could make changes to the installed templates as per their requirements.

rainboyan:
Package the templates into the deployed war? It dos’nt make any sense.
If need reuse the templates, I think custom profile is better.

rainboyan:
Ok, the scaffold templates may be used when using `static scaffold = Domain` in Controller, 

I found this:

```
if(!res?.exists()) {
                        def url = IOUtils.findResourceRelativeToClass(controllerClass.clazz, "/META-INF/templates/scaffolding/${shortViewName}.gsp")
                        res = url ? new UrlResource(url) : null
                    }
```

In my case, I could exclude the templates. Thank you.

Issue #104, opened 2022-04-11
Support set springBootVersion in Gradle


rainboyan:
When the new version of Spring Boot  was released, to align with Grails and Spring Boot, `grails-gradle-plugin` must need to be released too.

Like this:
```
grailsVersion=5.1.5
grailsGradlePluginVersion=5.1.3
springBootVersion=2.6.6
groovyVersion=3.0.7
gorm.version=7.2.1
```

PR #105, opened 2022-04-11
Support set springBootVersion in Gradle

rainboyan:
Fixed issue #104:

Update in `gradle.properties` will apply new version of Spring Boot

Use managed dependency versions that are provided by Spring Boot

`spring-framework.version`=5.3.19
`mysql.version`=8.0.28

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,6 +101,7 @@ dependencies {
         exclude group: "org.codehaus.gant", module: "gant_groovy1.8"
     }
 
+    compileOnly "org.apache.maven:maven-model:3.8.4"
     api "org.springframework.boot:spring-boot-gradle-plugin:$springBootVersion"
     api "io.spring.gradle:dependency-management-plugin:1.0.11.RELEASE"
     api "com.netflix.nebula:gradle-extra-configurations-plugin:7.0.0"
--- a/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -184,8 +184,12 @@ class GrailsGradlePlugin extends GroovyPlugin {
 
     @CompileDynamic
     private void applyBomImport(DependencyManagementExtension dme, project) {
+        String springBootVersion = project.findProperty('springBootVersion')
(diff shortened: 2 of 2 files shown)


rainboyan (commented):
@puneetbehl please review this feature.

puneetbehl:
I think we should allow changing spring version for Grails application. It should be tied to Grails version. Otherwise, it may create a lot of confusions and errors. I am trying to figure out an alternative to decouple this from Grails Gradle Plugin.

rainboyan:
@puneetbehl thank you for your time.

> I think we should allow changing spring version for Grails application.

Currently, we can change spring version via `spring.version`. but if the changed version is very low or high than the version tied to Grails version, there also will be errors. 
This PR allow change spring boot version, we also can check the version, just allow PATCH version. If there are incompatible changes, there may be the issues of Spring Boot or 3rd frameworks. For example, spring framework v5.3.18, logback v1.2.9, Spring boot and Grails users could change the version to upgrade or downgrade, it's always up to the users. 

> I am trying to figure out an alternative to decouple this from Grails Gradle Plugin.

Spring Boot Gradle Plugin is always published with Spring Boot, so can Grails. Maybe we should back the old days. Grails could be released a week later than Spring Boot, and patch releases are released once a month.

rainboyan:
@puneetbehl if `grails-shell` could contains `grails-bom.xml` in jar file, so **grails-grade-plugin** could read it and get the versions of Groovy, Micronaut, Spring Boot, this will keep the version same as Grails.
I have test this in my fork https://github.com/rainboyan/grails-gradle-plugin/commit/e2fedad69bd2ebb641dd34b7769c6bbd47bc929e https://github.com/rainboyan/grails-core/commit/1d26e1595c9500aea5d43c508600c3890187d27b, it works.

but in the `build.gradle` of app, we must include `grails-shell` in the classpath, because `grails-gradle-plugin` may use an old version.

```
buildscript {
    repositories {
        maven { url "https://repo.grails.org/grails/core" }
    }
    dependencies {
        classpath "org.grails:grails-gradle-plugin:$grailsGradlePluginVersion"
        classpath "org.grails:grails-shell:$grailsVersion"
        classpath "org.grails.plugins:hibernate5:7.2.2"
    }
}
```

PR #224, opened 2023-09-27
Fix version unspecified after update to Grails 6

matrei:
After update to Grails 6, setting the `version = x.y.z` in `build.gradle` no longer had effect. Setting it in `gradle.properties` did however. 

The problem seems to be that the `version` is (and must be) declared after the `plugins {}` section in build.gradle. This means that the version value is not available when the plugin code is run.

This pull request makes the code retrieving the version run after the project is evaluated, and fixes this issue.


matrei:
Fixes grails/grails-core#13011

puneetbehl (approved):

PR #244, opened 2023-10-27
Add source set dependencies back to `integrationTestImplementation`

### LINKED ISSUE #235: plugins can overwrite message bundles in integration tests
**OS:** macOS Sonoma 14.0 (23A344)
**JDK:** Eclipse Temurin JDK 17.0.6+10 (aarch64)
**Grails:** 6.0.0 installed through SDKman

The order of resources on the classpath from plugins seems to have changed between version 5.3.0 and 6.0.0. Before the project resources came first, but now the plugins come first.
This causes integration tests to load message bundles from the plugin, instead of the application's `grails-app/i18n/` folder.

Example project with reproduction here: https://github.com/matthijsbierman/grails6-i18n-bug

If you remove the dependency on the `mail` plugin, the test will pass. Downgrading to `implementation("org.grails:grails-gradle-plugin:5.3.0")` in `buildSrc/build.gradle` will also result in a passing test.

guillermocalvo:
After some investigation, I found out that the problem is not that "plugins can overwrite message bundles" in integration tests, but that version 6 of the plugin mistakenly removed main/test source set dependencies from configuration `integrationTestImplementation`.

I added back the _file dependencies_ so that it behaves like version 5. This solves #235

Please note that _configuration dependencies_ are the only ones that are deprecated and should be replaced by `Configuration#extendsFrom`.

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/core/IntegrationTestGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/core/IntegrationTestGradlePlugin.groovy
@@ -20,6 +20,7 @@ import groovy.transform.CompileDynamic
 import groovy.transform.CompileStatic
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.dsl.DependencyHandler
 import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.SourceSetContainer
 import org.gradle.api.tasks.SourceSetOutput
@@ -61,6 +62,9 @@ class IntegrationTestGradlePlugin implements Plugin<Project> {
             final File resources = new File(project.projectDir, "grails-app/conf")
             integrationTest.resources.srcDir(resources)
 
+            final DependencyHandler dependencies = project.dependencies
+            dependencies.add("integrationTestImplementation", mainSourceSetOutput)
+            dependencies.add("integrationTestImplementation", testSourceSetOutput)
             project.configurations.named("integrationTestImplementation") {
                 it.extendsFrom(project.configurations.named("testImplementation").get())
             }


puneetbehl:
This is removed intentionally as there was some deprecation message I saw which is something similar to following:

```
Adding a Configuration as a dependency is a confusing behavior which isn't recommended. This behaviour has been deprecated and is scheduled to be removed in Gradle 8.0. If you're interested in inheriting the dependencies from the Configuration you are adding, you should use Configuration#extendsFrom instead. 
```

I think you would be able to reproudce this by generating Gradle Build Scan against the local SNAPSHOT of this.

guillermocalvo:
@puneetbehl 

> This is removed intentionally as there was some deprecation message I saw which is something similar to following:
> 
> ```
> Adding a Configuration as a dependency is a confusing behavior which isn't recommended. This behaviour has been deprecated and is scheduled to be removed in Gradle 8.0. If you're interested in inheriting the dependencies from the Configuration you are adding, you should use Configuration#extendsFrom instead. 
> ```
> 
> I think you would be able to reproudce this by generating Gradle Build Scan against the local SNAPSHOT of this.

That's what I mean, **_configuration dependencies_ are the only ones that are deprecated**; _file dependencies_ are safe to keep.

See, you [replaced all dependencies in version 5](https://github.com/grails/grails-gradle-plugin/compare/5.3.x...6.1.x#diff-b2d64e138bcace184649885c6fc1a4b434afd8bbed9a0d50bb69f2d7503495e1R64):

```groovy
DependencyHandler dependencies = project.dependencies
dependencies.add("integrationTestImplementation", SourceSets.findMainSourceSet(project).output)
dependencies.add("integrationTestImplementation", SourceSets.findSourceSet(project, SourceSet.TEST_SOURCE_SET_NAME).output)
dependencies.add("integrationTestImplementation", project.configurations.findByName("testCompileClasspath"))
dependencies.add("integrationTestRuntimeOnly", project.configurations.findByName("testRuntimeClasspath"))
```

With:

```groovy
project.configurations.named("integrationTestImplementation") {
    it.extendsFrom(project.configurations.named("testImplementation").get())
}
project.configurations.named("integrationTestRuntimeOnly") {
    it.extendsFrom(project.configurations.named("testRuntimeOnly").get())
}
```

The two missing `SourceSet` dependencies are the root cause of #235:

```groovy
dependencies.add("integrationTestImplementation", SourceSets.findMainSourceSet(project).output)
dependencies.add("integrationTestImplementation", SourceSets.findSourceSet(project, SourceSet.TEST_SOURCE_SET_NAME).output)
```

These two don't produce any deprecation warnings.

puneetbehl:
Isn't the following doing the same thing:

https://github.com/grails/grails-gradle-plugin/blob/6.1.x/src/main/groovy/org/grails/gradle/plugin/core/IntegrationTestGradlePlugin.groovy#L53-L54

guillermocalvo:
> Isn't the following doing the same thing:
> 
> https://github.com/grails/grails-gradle-plugin/blob/6.1.x/src/main/groovy/org/grails/gradle/plugin/core/IntegrationTestGradlePlugin.groovy#L53-L54

Nope, that doesn't seem to work the way you would expect.

PR #250, opened 2023-11-03
Bump Grails version

guillermocalvo:
@puneetbehl I'm not sure if making this artifact depend on Grails `6.1.0-SNAPSHOT` is the way to do this, or I should use `6.1.0` (but then this PR will not build until `6.1.0` is released). What do you think? 🤔


puneetbehl:
It is fine to test with 6.1.0-SNAPSHOT, we need to make sure to update it to 6.1.0 before the release.

PR #273, opened 2024-01-16
fix(build): Cleanup dependency management

matrei:
All dependencies were declared with the api scope which is not necessary and which leaks dependencies to downstream projects. After this change it should no longer be necessary for Grails apps to force the version of groovy-xml. Several unused dependencies have also been removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -73,12 +73,16 @@ java {
 }
 
 configurations.configureEach {
+    // This is a workaround for grails-shell exposing a different version of Groovy
+    // than the one used by Gradle. This causes issues with the Groovy compiler.
     resolutionStrategy.force "org.codehaus.groovy:groovy-xml:$GroovySystem.version"
 }
 
 dependencies {
     implementation "io.github.gradle-nexus:publish-plugin:1.3.0"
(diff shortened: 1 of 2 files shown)


matrei:
@puneetbehl I have now tested this pull request on https://github.com/grails/grails-functional-tests by:
- Publishing this branch as `org.grails:grails-gradle-plugin:6.1.2-SNAPSHOT` to mavenLocal
- Adding mavenLocal() as repository to all the subprojects buildscript sections in `grails-functional-tests`
- Changing to `grailsGradlePluginVersion=6.1.2-SNAPSHOT` (was `6.1.0`) in `gradle.properties` in `grails-functional-tests`

And the build runs without issues.
(The only thing I needed to change was geb/selenium/webdriver versions/config to be able to run the integration test suite it on my computer.)

I have also previously successfully built https://github.com/grails/grails-spring-security-core (which also has several grails apps for testing) with these changes. In this project there is also code in `buildSrc/src/main/groovy` which would not compile without forcing `groovy-xml:3.0.13`. With the changes in this branch the forcing can be removed.

PR #289, opened 2024-03-15
Update to Grails 4.1.4

mattmoss:


mattmoss:
See: https://github.com/grails/grails-core/issues/13302#issuecomment-1990752960

Issue #305, opened 2024-07-18
7.x - Upgrade to Gradle 8 - set source/target compatibility to Java 17


codeconsole:
The [7.x branch](https://github.com/grails/grails-gradle-plugin/blob/7.0.x/build.gradle) should be running on Gradle 8.x with a min source/target compatibility of Java 17

https://docs.gradle.org/current/userguide/compatibility.html

https://github.com/orgs/grails/projects/3/views/2?pane=issue&itemId=71425178

codeconsole:
https://github.com/grails/grails-gradle-plugin/pull/290

codeconsole:
```
 Installing: gradle 8.5
Done installing!

Do you want gradle 8.5 to be set as default? (Y/n): n
sheiberg@Scotts-MBP-3 grails-gradle-plugin % gradle clean assemble 

> Task :compileGroovy FAILED
startup failed:
/Users/sheiberg/Documents/workspace/grails-gradle-plugin/src/main/groovy/org/grails/gradle/plugin/web/gsp/GspCompileOptions.groovy: 21: You cannot create an instance from the abstract class 'org.gradle.api.tasks.compile.GroovyForkOptions'.
 @ line 21, column 37.
       GroovyForkOptions forkOptions = new GroovyForkOptions()
                                       ^

1 error


FAILURE: Build failed with an exception.

```

codeconsole:
```
Installing: gradle 8.5
Done installing!

Do you want gradle 8.5 to be set as default? (Y/n): n
grails-gradle-plugin % gradle clean assemble 

> Task :compileGroovy FAILED
startup failed:
/grails-gradle-plugin/src/main/groovy/org/grails/gradle/plugin/web/gsp/GspCompileOptions.groovy: 21: You cannot create an instance from the abstract class 'org.gradle.api.tasks.compile.GroovyForkOptions'.
 @ line 21, column 37.
       GroovyForkOptions forkOptions = new GroovyForkOptions()
                                       ^

1 error


FAILURE: Build failed with an exception.
```

davydotcom:
GroovyForkOptions is now an abstract class i fixed this already but have a new error (not pushed yet)

```
> Unrecoverable compilation error: startup failed:
  General error during canonicalization: java.lang.NoClassDefFoundError: org.springframework.boot.loader.tools.LoaderImplementation
  
[...]

PR #307, opened 2024-07-25
Upgrading to Gradle 8.9

davydotcom:
This should work with gradle 8.9 but needs tested in a full grails app first. Work in Progress related to #305


davydotcom:
fyi the CI builds are only failing because the snapshot is not being updated from grails-core 7.0.x @sdelamo

codeconsole:
https://github.com/grails/grails-gradle-plugin/issues/314

PR #308, opened 2024-07-26
ci: Fix Github workflows for Grails 7

matrei:
The CI workflows were building with Java 11, 14 and 17, but Grails 7 is now built with Java 17 so we can no longer build with 11 and 14.


matrei:
The changelog workflows fails with auth issues, but that does not matter now.

PR #313, opened 2024-07-31
Revert "fix(core)!: Drop dependency `grails-shell`"

codeconsole:
Reverts grails/grails-gradle-plugin#287 necessary to get IntelliJ grails plugin working again.

Do we want to bring back the profile plugins as well or just add

```gradle
implementation "org.grails:grails-shell:$grailsVersion"
```

Are we deciding to bring back the old cli completely which could be run independently via `./grailsw` ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,6 +83,7 @@ dependencies {
     implementation "com.bmuschko:gradle-nexus-plugin:2.3.1"
     implementation "org.grails:grails-bootstrap:$grailsVersion"
     implementation "org.grails:grails-gradle-model:$grailsVersion"
+    implementation "org.grails:grails-shell:$grailsVersion"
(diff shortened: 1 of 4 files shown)


jamesfredley:
@codeconsole I found a reason to restore the `grails-shell` dependency here.  Without this revert when you add `org.grails.plugins:database-migration` to your `buildscript ` or `buildSrc` in order to run the `dbm-*` scripts, you end up having to manually add the `grails-shell` dependency to `buildscript ` or `buildSrc` also, which was not needed before this was removed.

The profiles plugin will be needed until grails-shell and profiles are no longer used.

codeconsole:
@jamesfredley so are you recommending we release this into 6.2.1?

jamesfredley:
@codeconsole Yes, I think we should release this into 6.2.1.

PR #321, opened 2024-08-21
update grails-shell to v7.0.0-SNAPSHOT

jamesfredley:


build.gradle
```diff
@@ -74,10 +74,20 @@ java {
     withSourcesJar()
 }
 
-configurations.configureEach {
-    // This is a workaround for grails-shell exposing a different version of Groovy
-    // than the one used by Gradle. This causes issues with the Groovy compiler.
-    resolutionStrategy.force "org.codehaus.groovy:groovy-xml:$GroovySystem.version"
+configurations.all {
```
matrei: Why do you choose `configurations.all` instead of `configurations.configureEach`?

build.gradle
```diff
@@ -74,10 +74,20 @@ java {
...
+    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        // This is a workaround for grails-shell exposing a different version of Groovy
+        // than the one used by Gradle. This causes issues with the Groovy compiler.
+        if (details.requested.group == 'org.apache.groovy') {
+            details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: "$GroovySystem.version")
```
matrei: GroovySystem.version returns a String, no need to use GString.

build.gradle
```diff
@@ -74,10 +74,20 @@ java {
...
+            details.because "Use Groovy 3.0.x library shipped with Gradle"
+        }
+
+        if (details.requested.group == 'org.codehaus.groovy') {
+            details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: "$GroovySystem.version")
```
matrei: GroovySystem.version returns a String, no need to use GString.

matrei (approved):

build.gradle
jamesfredley: It was just copied from another Grails project, I have adjusted it to configurations.configureEach which works well.

build.gradle
jamesfredley: Updated

build.gradle
jamesfredley: Updated

PR #139, opened 2022-08-14
Upgrade Apache Groovy to 4.0.2

puneetbehl:

PR #326, opened 2024-09-05
Groovy 4

codeconsole:
Remove large configurations block and set to groovy 4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,7 +74,8 @@ java {
     withSourcesJar()
 }
 
-configurations.all*.exclude group: 'org.apache.groovy', module: 'groovy-xml'
+// resolve conflict from Gradle's version of Groovy. Not needed if Gradle switches to 4.0
+configurations.configureEach { exclude group: 'org.apache.groovy', module: 'groovy-xml' }
 
 dependencies {
     implementation "org.apache.groovy:groovy-xml:$groovyVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -78,8 +78,6 @@ java {
 configurations.configureEach { exclude group: 'org.apache.groovy', module: 'groovy-xml' }
 
 dependencies {
-    implementation "org.apache.groovy:groovy-xml:$groovyVersion"
-    implementation "org.apache.groovy:groovy-templates:$groovyVersion"
     implementation "io.github.gradle-nexus:publish-plugin:2.0.0"
     implementation "com.bmuschko:gradle-nexus-plugin:2.3.1"
     implementation "org.grails:grails-bootstrap:$grailsVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -74,25 +74,11 @@ java {
     withSourcesJar()
 }
 
-configurations.configureEach {
-    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
-        // This is a workaround for grails-shell exposing a different version of Groovy
-        // than the one used by Gradle. This causes issues with the Groovy compiler.
-        if (details.requested.group == 'org.apache.groovy') {
-            details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: GroovySystem.version)
-            details.because "Use Groovy 3.0.x library shipped with Gradle"
-        }
-
-        if (details.requested.group == 'org.codehaus.groovy') {
-            details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: GroovySystem.version)
-            details.because "Use Groovy 3.0.x library shipped with Gradle"
-        }
-    }
-}
+configurations.all*.exclude group: 'org.apache.groovy', module: 'groovy-xml'
 
 dependencies {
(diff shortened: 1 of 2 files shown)


jamesfredley:
@codeconsole The only holdup on Groovy 4 for the grails-gradle-plugin is:

https://docs.gradle.org/current/userguide/compatibility.html#groovy "Gradle plugins written in Groovy must use Groovy 3.x for compatibility with Gradle and Groovy DSL build scripts."

jamesfredley:
Groovy 4 appears to work and eliminates the Groovy 3 vs 4 conflicts nicely, I just wish there was confirmation that the statement from Gradle is no longer valid.

build.gradle
```diff
@@ -74,25 +74,11 @@ java {
...
-            details.because "Use Groovy 3.0.x library shipped with Gradle"
-        }
-    }
-}
+configurations.all*.exclude group: 'org.apache.groovy', module: 'groovy-xml'
 
 dependencies {
-    implementation ( "org.codehaus.groovy:groovy-xml:$groovyVersion" )
-    implementation ( "org.codehaus.groovy:groovy-templates:$groovyVersion" )
+    implementation "org.apache.groovy:groovy-xml:$groovyVersion"
```
jamesfredley: I believe these lines can be removed entirely.

```
implementation "org.apache.groovy:groovy-xml:$groovyVersion"
implementation "org.apache.groovy:groovy-templates:$groovyVersion"
```

build.gradle
```diff
@@ -74,25 +74,11 @@ java {
...
-            details.because "Use Groovy 3.0.x library shipped with Gradle"
-        }
-
-        if (details.requested.group == 'org.codehaus.groovy') {
-            details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: GroovySystem.version)
-            details.because "Use Groovy 3.0.x library shipped with Gradle"
-        }
-    }
-}
+configurations.all*.exclude group: 'org.apache.groovy', module: 'groovy-xml'
```
matrei: I find It's better to use `configurations.configureEach` as `configurations.all` will configure all configurations eagerly (even unused configurations), while the former does so only lazily upon usage. I also like to write a comment as to why, I for example exclude something, because then it's really easy for the next developer reading the file to understand.

matrei:
> Groovy 4 appears to work and eliminates the Groovy 3 vs 4 conflicts nicely, I just wish there was confirmation that the statement from Gradle is no longer valid.

@codeconsole Have you tried to use the plugins from this project compiled with Groovy 4?

jamesfredley:
@puneetbehl submitted a PR with the Groovy 4 update, about 2 years ago, so this appears to have been planned for Grails 6, but Groovy 4 got bumped to Grails 7.

https://github.com/grails/grails-gradle-plugin/pull/139

codeconsole:
> > Groovy 4 appears to work and eliminates the Groovy 3 vs 4 conflicts nicely, I just wish there was confirmation that the statement from Gradle is no longer valid.
> 
> @codeconsole Have you tried to use the plugins from this project compiled with Groovy 4?

`:compileGroovyPages`  works, but `:urlMappingsReport` fails

```
./gradlew :urlMappingsReport

> Task :urlMappingsReport FAILED
Context failed to load: Bean definition [io.micronaut.spring.context.factory.ImportBeanRegistrarProcessor] could not be loaded: Could not initialize class io.micronaut.inject.annotation.AnnotationMetadataSupport


```

jamesfredley:
I tested it against `grails-database-migration` and `gorm-mongodb`: `assemble`, `build `& `groovydoc `- no issues

and it allows removal of the following from buildScript:

```
configurations.configureEach {
        resolutionStrategy.eachDependency { DependencyResolveDetails details ->
            // This is a workaround for dependencies exposing a different version of Groovy
            // than the one used by Gradle. This causes issues with the Groovy compiler.
            if (details.requested.group == 'org.apache.groovy' || details.requested.group == 'org.codehaus.groovy') {
                details.useTarget(group: "org.codehaus.groovy", name: details.requested.name, version: GroovySystem.version)
                details.because "Use Groovy 3.0.x library shipped with Gradle"
            }
        }
    }
```

There are still some 7.0.0-SNAPSHOT dependencies that are referencing org.codehaus.groovy which separately need updated.

org.grails:grails-gsp:7.0.0-SNAPSHOT
org.grails:grails-web-gsp-taglib:7.0.0-SNAPSHOT
org.grails:grails-web-gsp:7.0.0-SNAPSHOT
org.grails:grails-web-jsp:7.0.0-SNAPSHOT

PR for these https://github.com/grails/grails-gsp/pull/477

jamesfredley:
The following should resolve `Context failed to load: Bean definition [io.micronaut.spring.context.factory.ImportBeanRegistrarProcessor] could not be loaded: Could not initialize class io.micronaut.inject.annotation.AnnotationMetadataSupport`

```
"io.micronaut:micronaut-inject-groovy:$micronautVersion"
"io.micronaut:micronaut-inject:$micronautVersion"
```

jamesfredley (approved):

PR #331, opened 2024-09-16
Gradle 8 compatability: compileWebappGroovyPages depends on copyAstClasses

codeconsole:


jamesfredley (approved):

codeconsole:
@jamesfredley this is breaking

```
org.gradle.api.plugins.InvalidPluginException: An exception occurred applying plugin request [id: 'org.grails.grails-gsp', version: '7.0.0-SNAPSHOT']
    at org.gradle.plugin.use.internal.DefaultPluginRequestApplicator.exceptionOccurred(DefaultPluginRequestApplicator.java:179)

An exception occurred applying plugin request [id: 'org.grails.grails-gsp', version: '7.0.0-SNAPSHOT']
> Failed to apply plugin 'org.grails.grails-gsp'.
   > Cannot read the array length because "<local2>" is null
```

but adding this directly  in the app `build.gradle` fixes

```
compileWebappGroovyPages.dependsOn(copyAstClasses)
```

```
* What went wrong:
A problem was found with the configuration of task ':compileWebappGroovyPages' (type 'GroovyPageForkCompileTask').
  - Gradle detected a problem with the following location: '/Users/sheiberg/Documents/workspace/grails-plugin-sitemesh3/build/classes/groovy/main'.
    
    Reason: Task ':compileWebappGroovyPages' uses this output of task ':copyAstClasses' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':copyAstClasses' as an input of ':compileWebappGroovyPages'.
      2. Declare an explicit dependency on ':copyAstClasses' from ':compileWebappGroovyPages' using Task#dependsOn.
      3. Declare an explicit dependency on ':copyAstClasses' from ':compileWebappGroovyPages' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/8.10.1/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.

```


@matrei you have any thoughts on what causes the above exception and how to fix it directly in the plugin?

jamesfredley:
@codeconsole give https://github.com/grails/grails-gradle-plugin/pull/332 a try.  This one was more complicated than normal since the two tasks did not always exist at the same time and required a try/catch.

PR #334, opened 2024-09-17
compile grails-gradle-plugin with the Groovy version provided by Gradle

jamesfredley:
to ensure build compatibility with Gradle, currently Groovy 3.0.x see: https://docs.gradle.org/current/userguide/compatibility.html#groovy

This will prevent running into Groovy 4 vs 3 compatibility issues like:  https://github.com/bertramdev/asset-pipeline/pull/350, when Gradle executes the build using it's embedded Groovy Version, currently 3.0.22.

This will also be done on `grails-gradle-model`, `grails-shell` and `grails-bootstrap` in `grails-core`.   https://github.com/grails/grails-core/pull/13653

Move all versions numbers to gradle.properties and update dependencies to latest compatible version

Simplify `ext."signing.*"`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,11 @@ java {
 configurations.configureEach { exclude group: 'org.apache.groovy' }
 
 dependencies {
+    // compile grails-gradle-plugin with the Groovy version provided by Gradle
+    // to ensure build compatibility with Gradle, currently Groovy 3.0.x
+    // see: https://docs.gradle.org/current/userguide/compatibility.html#groovy
+    compileOnly    "org.codehaus.groovy:groovy:$GroovySystem.version"
+
     implementation "io.github.gradle-nexus:publish-plugin:2.0.0"
     implementation "com.bmuschko:gradle-nexus-plugin:2.3.1"
     implementation "org.grails:grails-bootstrap:$grailsVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,11 @@ java {
 configurations.configureEach { exclude group: 'org.apache.groovy', module: 'groovy-xml' }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


matrei:
Is this a safe-guard for future changes to the build?
Because, as far as I can tell Groovy 4 is not on the `compileClasspath` right now.

```console
./gradlew dI --dependency groovy --configuration compileClasspath

> Task :dependencyInsight
No dependencies matching given input were found in configuration ':compileClasspath'
```

> This would also need to be done on grails-bootstrap, grails-gradle-model and grails-shell.

These however, do have Groovy 4 on the `compileClasspath`, and `grails-shell` in particular also has `org.apache.groovy:groovy-ant`, `org.apache.groovy:groovy-json`, `org.apache.groovy:groovy-jmx` and `org.apache.groovy:groovy-xml` in `api` scope which, I believe, is why we need this in the `grails-gradle-plugin` `build.gradle`:
```groovy
// resolve conflict from Gradle's version of Groovy. Not needed if Gradle switches to 4.0
configurations.configureEach { exclude group: 'org.apache.groovy' }

```

jamesfredley:
Yes, this change plus similar compile changes on `grails-bootstrap`, `grails-gradle-model` and `grails-shell` will be a safeguard against future changes that could result in a release that breaks build and is not caught until after release.   This change can/will be limited to compileOnly for these 4 projects. 

Recent changes to grails-gradle-plugin had us move to groovy 4 and then back to 3.   The need to ensure Groovy 3 is used to compile was illustrated by the asset-pipeline-gradle:5.0.0 release and it seemed like a good idea to do the same in Grails.

matrei (approved):

puneetbehl (approved):

PR #337, opened 2024-09-21
Groovy 4 for projects, Groovy 3 for Gradle plugins

codeconsole:
If `groovyVersion` is set to 3.x, forces groovy 3.  if `groovyVersion is set to 4.x, forces groovy 4.

No need for adding a bunch one configure code to every project.

FYI @jamesfredley  the groovy 3 version used by gradle is whatever groovy version that is set by the project. Setting the version to `GroovySystem.version` will not be the same for every project unless it is explicitly set using `groovyVersion`

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -261,22 +261,33 @@ class GrailsGradlePlugin extends GroovyPlugin {
     protected void configureGroovy(Project project) {
         final String groovyVersion = project.properties['groovyVersion']
         if (groovyVersion) {
-            project.configurations.all({ Configuration configuration ->
-                configuration.resolutionStrategy.eachDependency({ DependencyResolveDetails details ->
-                    String dependencyName = details.requested.name
-                    String group = details.requested.group
-                    if (group == 'org.codehaus.groovy') {
-                        if (dependencyName == 'groovy-all') {
-                            details.useTarget "org.apache.groovy:groovy:$groovyVersion"
-                        } else {
-                            details.useTarget "org.apache.groovy:$dependencyName:$groovyVersion"
+            if (groovyVersion.startsWith("4.")) {
+                project.configurations.configureEach { Configuration configuration ->
+                    configuration.resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+                        String dependencyName = details.requested.name
+                        String group = details.requested.group
+                        if (group == 'org.codehaus.groovy') {
+                            if (dependencyName == 'groovy-all') {
+                                details.useTarget "org.apache.groovy:groovy:$groovyVersion"
+                            } else {
+                                details.useTarget "org.apache.groovy:$dependencyName:$groovyVersion"
+                            }
+                            details.because("Groovy version substituted for Groovy $groovyVersion")
+                        } else if (group == 'org.apache.groovy' && dependencyName.startsWith('groovy')) {
+                            details.useVersion(groovyVersion)
                         }
-                        details.because('groovy version substituted for groovy 4')
-                    } else if (group == 'org.apache.groovy' && dependencyName.startsWith('groovy')) {
-                        details.useVersion(groovyVersion)
                     }
-                } as Action<DependencyResolveDetails>)
-            } as Action<Configuration>)
+                }
+            } else if (groovyVersion.startsWith("3.")) {
+                project.configurations.configureEach { Configuration configuration ->
+                    configuration.resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+                        if ((details.requested.group == 'org.codehaus.groovy' || details.requested.group == 'org.apache.groovy') && details.requested.name != 'groovy-bom') {
+                            details.useTarget(group: 'org.codehaus.groovy', name: details.requested.name, version: "$GroovySystem.version")
+                            details.because "Use Groovy version $GroovySystem.version provided by Gradle"
+                        }
+                    }
+                }
+            }
         }
     }
 


codeconsole:
So you don't have to add thing like  https://github.com/gpc/fields/pull/356

```gradle
    configurations.configureEach {
        resolutionStrategy.eachDependency { DependencyResolveDetails details ->
            if ((details.requested.group == 'org.codehaus.groovy' || details.requested.group == 'org.apache.groovy') && details.requested.name != 'groovy-bom') {
                details.useTarget(group: 'org.codehaus.groovy', name: details.requested.name, version: "$GroovySystem.version")
                details.because "Use version of Groovy provided by Gradle"
            }
        }
    }
```

if `grails-gradle-plugin` is used by the project

If we are trying to keep all our projects coordinated, we should try and eliminate these configurations blocks in all of them.

matrei:
> the groovy 3 version used by gradle is whatever groovy version that is set by the project. Setting the version to GroovySystem.version will not be the same for every project unless it is explicitly set using groovyVersion

I might be misunderstanding you here, but I'm not sure this statement is correct. Try the following gradle task:
```groovy
tasks.register('printGroovyVersions') {
    doLast {
        println "Project Groovy version: ${project.findProperty('groovyVersion')}"
        println "Gradle Groovy version: $GroovySystem.version"
    }
}
```

jamesfredley:
@codeconsole I love centralizing the config in the grails-gradle-plugin, and that is the predominant case. But there are some exceptions that will need to be supported.

This may need some modifications for the following, I'll figure that out as I get that work for grails-core wrapped up and then test this against it.  

Gradle still using 3.0.x presents the need for projects to have 3.0.x used for compileOnly on Gradle Plugins and dependencies used by Gradle Plugins, while groovyVersion is set to 4.0.x, such as in grails-core.  And in cases like grails-shell and grails-boot in grails-core, after it is compiled with 3.0.x, we then test it with 4.0.x to make sure it is cross version compatible, since it will be run in 3.0.x by Gradle during build in other projects and 4.0.x when the Grails project runs tasks other than build. 

I really wish Gradle was on 4.0.x, because this work to compile with Gradle's version seems unuseful, but needs to happen to prevent releases that don't run during the build.  

And then buildSrc and buildscript{} are being run in Groovy 3.0.x by gradle, even when dependencies have been forced to org.apache.groovy with 4.0.x.  That works fine as long as the package in the code exists in the org.apache.groovy jars, which still includes several org.codehause.groovy packages in addition to org.apache.groovy packages.

jamesfredley:
We have been able to remove most of the version override blocks with the current grails-gradle-plugin.  For example there are none in my fork of your website demo for Grails 7, since we have migrated all of the dependencies to 4. 

Projects with dependencies like `io.github.groovylang.groovydoc:groovydoc-gradle-plugin:1.0.1` are a current exception, because it is on groovy 2.x.

jamesfredley:
grails-core is not using the grails-gradle-plugin during build, so this change will not impact the work in https://github.com/grails/grails-core/pull/13653.

codeconsole:
> > the groovy 3 version used by gradle is whatever groovy version that is set by the project. Setting the version to GroovySystem.version will not be the same for every project unless it is explicitly set using groovyVersion
> 
> I might be misunderstanding you here, but I'm not sure this statement is correct. Try the following gradle task:
> 
> ```groovy
> tasks.register('printGroovyVersions') {
>     doLast {
>         println "Project Groovy version: ${project.findProperty('groovyVersion')}"
>         println "Gradle Groovy version: $GroovySystem.version"
>     }
> }
> ```

@matrei  This was the behavior I observed, perhaps coincidental when I set groovyVersion to `3.0.22` in the fields plugin?  Is the default groovy for 8.10.1 really 3.0.22??
```
./gradlew --version

------------------------------------------------------------
Gradle 8.10.1
------------------------------------------------------------

Build time:    2024-09-09 07:42:56 UTC
Revision:      8716158d3ec8c59e38f87a67f1f311f297b79576

Kotlin:        1.9.24
Groovy:        3.0.22
Ant:           Apache Ant(TM) version 1.10.14 compiled on August 16 2023
Launcher JVM:  21.0.2 (Oracle Corporation 21.0.2+13-58)
``` 

I assumed that since Gradle can't upgrade to 4, they most likely can not keep the Gradle Groovy version up to date.

Either way, this code works by setting any Groovy 3 version to `GroovySystem.version` if `groovyVersion` is set which is the intention of the code.

codeconsole:
@jamesfredley this code only works if `groovyVersion` is set.  If it is not set, it does nothing.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -261,22 +261,33 @@ class GrailsGradlePlugin extends GroovyPlugin {
...
-            project.configurations.all({ Configuration configuration ->
-                configuration.resolutionStrategy.eachDependency({ DependencyResolveDetails details ->
-                    String dependencyName = details.requested.name
-                    String group = details.requested.group
-                    if (group == 'org.codehaus.groovy') {
-                        if (dependencyName == 'groovy-all') {
-                            details.useTarget "org.apache.groovy:groovy:$groovyVersion"
-                        } else {
-                            details.useTarget "org.apache.groovy:$dependencyName:$groovyVersion"
+            if (groovyVersion.startsWith("4.")) {
```
jamesfredley, line 264: I think we should flip this if/else so that it works without change when we go to Groovy 5.  Something like this:

```

            if (groovyVersion.startsWith("3.")) {
                project.configurations.configureEach { Configuration configuration ->
                    configuration.resolutionStrategy.eachDependency { DependencyResolveDetails details ->
                        if ((details.requested.group == 'org.codehaus.groovy' || details.requested.group == 'org.apache.groovy') && details.requested.name != 'groovy-bom') {
                            details.useTarget(group: 'org.codehaus.groovy', name: details.requested.name, version: "$GroovySystem.version")
                            details.because "Use Groovy version $GroovySystem.version provided by Gradle"
                        }
                    }
                }
            } else {
                project.configurations.configureEach { Configuration configuration ->
                    configuration.resolutionStrategy.eachDependency { DependencyResolveDetails details ->
                        String dependencyName = details.requested.name
                        String group = details.requested.group
                        if (group == 'org.codehaus.groovy') {
                            if (dependencyName == 'groovy-all') {
                                details.useTarget "org.apache.groovy:groovy:$groovyVersion"
                            } else {
                                details.useTarget "org.apache.groovy:$dependencyName:$groovyVersion"
                            }
                            details.because("Groovy version substituted for Groovy $groovyVersion")
                        } else if (group == 'org.apache.groovy' && dependencyName.startsWith('groovy')) {
                            details.useVersion(groovyVersion)
                        }
                    }
                }
            }
```

jamesfredley:
I've tested this on grails-profile:base, gorm-hibernate5 and fields.   On fields it is not yet possible to remove the groovy substitution from buildscript{}.  

```
A problem occurred configuring root project 'fields'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve org.codehaus.groovy:groovy-groovydoc:2.4.6.
     Required by:
         root project : > io.github.groovylang.groovydoc:groovydoc-gradle-plugin:1.0.1
      > Module 'org.codehaus.groovy:groovy-groovydoc' has been rejected:
           Cannot select module with conflict on capability 'org.codehaus.groovy:groovy-groovydoc:2.4.6' also provided by [org.apache.groovy:groovy-groovydoc:4.0.22(groovyRuntimeElements)]
```

codeconsole:
> I've tested this on grails-profile:base, gorm-hibernate5 and fields. On fields it is not yet possible to remove the groovy substitution from buildscript{}.
> 
> ```
> A problem occurred configuring root project 'fields'.
> > Could not resolve all artifacts for configuration ':classpath'.
>    > Could not resolve org.codehaus.groovy:groovy-groovydoc:2.4.6.
>      Required by:
>          root project : > io.github.groovylang.groovydoc:groovydoc-gradle-plugin:1.0.1
>       > Module 'org.codehaus.groovy:groovy-groovydoc' has been rejected:
>            Cannot select module with conflict on capability 'org.codehaus.groovy:groovy-groovydoc:2.4.6' also provided by [org.apache.groovy:groovy-groovydoc:4.0.22(groovyRuntimeElements)]
> ```

@jamesfredley 
You have to set the groovyVersion in gradle.properties. Otherwise it does nothing. I tested it on fields and it worked.

jamesfredley (approved):
@codeconsole Yes, `groovyVersion=4.0.23` on the` 6.0.x` branch.  buildscrip{} configuration does not get modified by `project.configurations.configureEach`, which is the reason we have had to do version substitution separately in both locations and also in buildSrc, when used.  

The need for Groovy version substitution is almost entirely gone with the last change in configureGroovy(Project project) and this change, plus migrating the bulk of dependencies to Groovy 4.

I've adjusted https://github.com/gpc/fields/pull/356 to exclude org.codehaus.groovy from that old dependency.

I do think we should still flip that **if/else** so this works when we switch to Groovy 5 and this is ready to merge in addition to https://github.com/gpc/fields/pull/356

PR #346, opened 2024-11-03
Default import of java.time.*

### LINKED ISSUE #345: java.time needs to be better supported in Grails 7
We can start with a compiler configuration pending the outcome of this [issue](https://issues.apache.org/jira/projects/GROOVY/issues/GROOVY-11513)

Need a smooth pathway from legacy java.util.Date that does not requiring adding imports all over the place.

https://groovy.apache.org/blog/groovy-dates-and-times-cheat

Can be turned off using 

```gradle
grails {
    importJavaTime = false
}
```

Groovy [currently star imports](https://github.com/apache/groovy/blob/d557f2729ec00865455f47c30eba9b3d03c25aea/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java#L99)
```"java.lang.", "java.util.", "java.io.", "java.net.", "groovy.lang.", "groovy.util."```

Results in the following automatically imported:
```
Clock
DateTimeException
DayOfWeek
Duration
Instant
InstantSource
LocalDate
LocalDateTime
LocalTime
Month
MonthDay
OffsetDateTime
OffsetTime
Period
Year
YearMonth
ZonedDateTime
ZoneId
ZoneOffset
```

### PR DESCRIPTION
**Pull request by codeconsole**
Enables smooth transition from legacy java.util.Date to [java.time](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/package-summary.html)

Fixes https://github.com/grails/grails-gradle-plugin/issues/345

### CODE REVIEW

File: src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy [hunk 1]
```diff
@@ -43,6 +43,11 @@ class GrailsExtension {
      */
     boolean pathingJar = false
 
+    /**
+     * Whether java.time.* package should be a default import package
+     */
+    boolean javaTime = true
```
matrei: How about naming the property something more descriptive? `autoImportDateTimeApi`?

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -652,6 +654,31 @@ class GrailsGradlePlugin extends GroovyPlugin {
         output?.classesDirs ?: project.files(new File(project.buildDir, "classes/main"))
     }
 
+    protected void configureJavaTime(Project project) {
```
matrei: Is `autoImportDateTimeApi(Project project)` a more descriptive name?

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -142,6 +142,8 @@ class GrailsGradlePlugin extends GroovyPlugin {
 
         configureRunCommand(project)
 
+        configureJavaTime(project)
```
matrei: Could we use `configureGroovy(project)` and invoke `autoImportDateTimeApi(Project project)` from there?

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -652,6 +654,31 @@ class GrailsGradlePlugin extends GroovyPlugin {
         output?.classesDirs ?: project.files(new File(project.buildDir, "classes/main"))
     }
 
+    protected void configureJavaTime(Project project) {
+        GrailsExtension grailsExt = project.extensions.getByType(GrailsExtension)
+
+        project.tasks.withType(GroovyCompile).configureEach { groovyCompileTask ->
+            if (grailsExt.javaTime) {
+                groovyCompileTask.doFirst {
+                    def configScriptStream = getClass().getResourceAsStream("/GrailsCompilerConfig.groovy")
```
matrei: Could we not just add the code here as:
```groovy
def configScript = '''
def dateTimeApiImportCustomizer = new org.codehaus.groovy.control.customizers.ImportCustomizer()
dateTimeApiImportCustomizer.addStarImports('java.time')

configuration.addCompilationCustomizers(dateTimeApiImportCustomizer)
'''
```
I'm thinking, if there are more customizers we want to add in the future, this will allow for that.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -652,6 +654,31 @@ class GrailsGradlePlugin extends GroovyPlugin {
         output?.classesDirs ?: project.files(new File(project.buildDir, "classes/main"))
     }
 
+    protected void configureJavaTime(Project project) {
+        GrailsExtension grailsExt = project.extensions.getByType(GrailsExtension)
+
+        project.tasks.withType(GroovyCompile).configureEach { groovyCompileTask ->
+            if (grailsExt.javaTime) {
```
matrei: Move this condition to `configureGroovy(Project project)` to conditionally invoke `autoImportDateTimeApi(Project project)`?

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jeffscottbrown: I think this isn't just auto importing DateTime.  I think it is importing `java.time.*`.  Is that the case?  If it is, I don't think `autoImportDateTimeApi` is a good name.

matrei: Well, `autoImportDateTimeApi` is at least the sentiment of what we are trying to do. There are of course a couple of sub-packages that also belong to the DateTime Api.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: Couldn’t you just add them to the file? Having a file you could select it with the ide.

matrei: Well, if you have them all in the same file, you can't turn them on or off individually.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
jeffscottbrown: I understand now.  Thank you!

codeconsole: @matrei my sentiment was always just import java.time.* and that is why I named it JavaTime.  Are you suggesting we should import everything else? 

```
java.time.chrono.*
java.time.format.*
java.time.temporal.*
java.time.zone.*
```

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
codeconsole: @matrei Paul [refers](https://groovy.apache.org/blog/groovy-dates-and-times-cheat) to it is as java.time in the groovy documentation

codeconsole: All the other names in there are pretty succinct.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: true, but couldn't you also just append more files as I am doing right now if a configScript is currently set?  I was hoping you would be able to reference the config script file in IntelliJ, but I have yet to see it work.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: @matrei done, but I think we are getting close to not even needing that groovy substitution logic.

matrei: But still, we are configuring groovy. Even if we remove the substitutions, we can still keep the method.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
matrei: >  Are you suggesting we should import everything else?

I don't know. Are there any drawbacks?

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
matrei: >  Paul [refers](https://groovy.apache.org/blog/groovy-dates-and-times-cheat) to it is as java.time in the groovy documentation

I would guess he is using the package name. He is also using classes from the subpackages when referencing `java.time`. 

There is also this (which I think is the more common name):
> Since Java 8, the JDK has included the [JSR-310](https://jcp.org/en/jsr/detail?id=310) Date Time API. We refer to these classes as the new date classes.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
```diff
@@ -654,31 +674,6 @@ class GrailsGradlePlugin extends GroovyPlugin {
         output?.classesDirs ?: project.files(new File(project.buildDir, "classes/main"))
     }
 
-    protected void configureJavaTime(Project project) {
```
matrei: Actually, I meant to keep this method (although with a different name), and call it from the configureGroovy method.

```groovy
protected void configureGroovy(Project project) {
     // some other stuff

    if (grailsExt.autoImportDateTimeApi) {
        autoImportDateTimeApi(project)
    }
}
```

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
```diff
@@ -46,7 +46,7 @@ class GrailsExtension {
     /**
      * Whether java.time.* package should be a default import package
      */
-    boolean javaTime = true
+    boolean importJavaTime = true
```
matrei, line 49: We are getting there 😄. I can live with this.

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
matrei: > All the other names in there are pretty succinct.

I know...

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: not sure if it is necessary or needed.

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
codeconsole: @matrei  so importJavaTime ok or do you want importDateTime or  importDateTimeApi?
I think autoImportDateTimeApi is too long compared to other properties.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: done

src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
codeconsole: ok, lol, but if you really really want something else, I can change. I really don't care, just prefer short.

src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
codeconsole: the groovy examples only needed: 

```groovy
import java.time.*
import static java.time.temporal.ChronoField.*
import java.time.temporal.ChronoUnit
import java.time.format.FormatStyle
```
to work

codeconsole: @matrei drawback is that I am not a fan of start imports to begin with, but I feel we need to offer a smooth painless transition away from the java Date class.

codeconsole:
I am going to merge this.  I am not opposed to any changes.  I just want to keep things moving.  The discussion can continue and this can continue to evolve, but there are other areas right now where this needs to be expanded on and the work can not continue without the merge.   For instance, LocalDateTime currently does not render an input.   Let's continue discussion in #345

PR #347, opened 2024-11-03
Remove `-plain` suffix from Grails Plugin artifacts

matrei:


src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
```diff
@@ -186,13 +187,14 @@ class GrailsPluginGradlePlugin extends GrailsGradlePlugin {
...
-        jarTask.exclude "application.yml"
-        jarTask.exclude "application.groovy"
-        jarTask.exclude "logback.groovy"
-        jarTask.exclude "logback.xml"
+        project.tasks.named(SpringBootPlugin.BOOT_JAR_TASK_NAME) {
+            it.enabled = false // Grails Plugins should not produce a bootJar
+        }
+        project.tasks.named(JavaPlugin.JAR_TASK_NAME, Jar) {
+            it.enabled = true
+            it.archiveClassifier.set('') // Remove '-plain' suffix from jar file name
```
jdaugherty, line 195: So this will impact any grails plugin (which is probably a very good thing) - @jamesfredley started a google document to gather upgrade gotchas.  @codeconsole moved that document to here: https://github.com/grails/grails-core/blob/7.0.x/UPGRADE7.md  Can you add a bullet point documenting this change for plugin authors?

jdaugherty (approved):
Thrilled to see this change.  Thank you!

codeconsole (approved):

src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy
matrei: > So this will impact any grails plugin

Yes, any project that applies the `org.grails.grails-plugin` gradle plugin will be impacted.

> Can you add a bullet point documenting this change for plugin authors?

Here is an PR for the Grails 7 gotchas: https://github.com/grails/grails-core/pull/13825

PR #355, opened 2024-11-20
#350 - Rework publishing plugins so they can be used externally and on multi-project builds

### LINKED ISSUE #350: Improve the Gradle Publish Plugins & Move them to public available instead of internal
Improve the `Internal` Grails Gradle Publish Plugins so that they can be used in sub projects & any Grails organization customizations can be externalized.

jdaugherty:
This change resolves #350.  At a high level it: 
1. Removes unused extension properties
2. Removes "internal" from publish namespaces & Rename plugins to reflect they aren't only for grails central
3. Updates the gradle enterprise logic that was commented out with the initial upgrade to grails 7 (jumps to latest version)
4. Updates the github actions to match grails-core (to pickup the gradle enterprise updates)
5. Adds documentation for the publish plugin
6. Supports publishing to artifactory or maven central for snapshots
7. Migrates the username, password, and url properties to properties / environment variables only.  using a publish plugin will require a snapshot url property to be defined based on the snapshot repository type.  My assumption is we'll add these to the gradle.properties in each project
8. Supports defining the `grailsPublish` block at the subproject level.

With these changes, any project can remove the gradle boiler plate of publishing to maven central & an artifactory snapshot instance.  Technically, it doesn't have to be a grails project to use this plugin.  If it is a grails plugin then the plugin.xml will be added as an artifact for the publish plugin and for the profile plugin it will generate the associated profile.xml.

I tested the snapshot portion of this logic with this code [spring security rest code](https://github.com/jdaugherty/grails-spring-security-rest/commit/ed5d25e99d53d6e22f13af22e25ac72ded130c0f).  @jeffscottbrown was kind enough to assist in testing.  You can find an example artifact [here](https://repo.grails.org/ui/native/plugins3-snapshots-local/org/grails/plugins/spring-security-rest-testapp-profile/6.0.1-SNAPSHOT/ ) - (I published to the wrong repo initially, oops).

Once this change is merged, all of the repositories with `grailsPublish` will have broken publishing since the urls will be missing. I plan to go through all of them quickly and set the appropriate snapshot url (plugins [go here](https://repo.grails.org/grails/plugins3-snapshots-local) & profiles [go here](https://repo.grails.org/grails/libs-snapshots-local)) to fix this breakage.  Separately, I'll go through the various multi-projects that couldn't use grailsPublish and switch them to grails publish (like the spring security rest plugin).  

As part of the milestone announcement, we can offer these plugins to people so they can use them to simplify plugin upgrades.  I'd like to provide an example github action too.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: ['17']
+        java: ['17', '21']
     env:
       WORKSPACE: ${{ github.workspace }}
     steps:
@@ -67,8 +67,8 @@ jobs:
     permissions:
       contents: write
     env:
-      GIT_USER_NAME: 'grails-build'
-      GIT_USER_EMAIL: 'grails-build@users.noreply.github.com'
+      GIT_USER_NAME: grails-build
+      GIT_USER_EMAIL: grails-build@users.noreply.github.com
     steps:
       - name: "📥 Checkout repository"
         uses: actions/checkout@v4
@@ -81,15 +81,13 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
         with:
           develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
-      - name: Generate Groovydoc
+      - name: "📜 Generate Groovydoc"
         id: groovydoc
-        uses: gradle/gradle-build-action@v3
         env:
           DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
           DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
-        with:
-          arguments: groovydoc
-      - name: Publish to Github Pages
+        run: ./gradle groovydoc
+      - name: "🚀 Publish to Github Pages"
         id: docs
         if: success()
         uses: grails/github-pages-deploy-action@v2
@@ -100,6 +98,5 @@ jobs:
           BRANCH: gh-pages
           FOLDER: build/docs
           DOC_FOLDER: gh-pages
-          COMMIT_EMAIL: behlp@unityfoundation.io
-          COMMIT_NAME: Puneet Behl
-          VERSION: ${{ needs.publish.outputs.release_version }}
+          COMMIT_EMAIL: ${{ env.GIT_USER_EMAIL }}
+          COMMIT_NAME: ${{ env.GIT_USER_NAME }}
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -67,8 +67,8 @@ jobs:
     permissions:
       contents: write
     env:
-      GIT_USER_NAME: puneetbehl
-      GIT_USER_EMAIL: behlp@unityfoundation.io
+      GIT_USER_NAME: 'grails-build'
+      GIT_USER_EMAIL: 'grails-build@users.noreply.github.com'
     steps:
       - name: "📥 Checkout repository"
         uses: actions/checkout@v4
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -13,8 +13,6 @@ jobs:
     strategy:
       matrix:
         java: ['17', '21']
-    env:
-      WORKSPACE: ${{ github.workspace }}
     steps:
       - name: "📥 Checkout repository"
         uses: actions/checkout@v4
@@ -86,7 +84,7 @@ jobs:
         env:
           DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
           DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
-        run: ./gradle groovydoc
+        run: ./gradlew groovydoc
       - name: "🚀 Publish to Github Pages"
         id: docs
         if: success()
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -16,41 +16,50 @@ jobs:
     env:
       WORKSPACE: ${{ github.workspace }}
     steps:
-      - uses: actions/checkout@v4
-      - name: Set up JDK
+      - name: "📥 Checkout repository"
+        uses: actions/checkout@v4
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: ${{ matrix.java }}
-      - name: Run Build
-        id: build
-        uses: gradle/gradle-build-action@v3
-        env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
         with:
-          arguments: build -Dgeb.env=chromeHeadless
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+      - name: "🔨 Build project"
+        id: build
+        env:
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: ./gradlew build
   publish:
     if: github.event_name == 'push'
     needs: ['build']
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - name: Set up JDK
+      - name: "📥 Checkout repository"
+        uses: actions/checkout@v4
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: 17
-      - name: Publish Artifacts (repo.grails.org)
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+      - name: "📤 Publish Snapshot Artifacts to Artifactory (repo.grails.org/libs-snapshot-local)"
         id: publish
-        uses: gradle/gradle-build-action@v3
         env:
           ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
-        with:
-          arguments: | 
-            -Dorg.gradle.internal.publish.checksums.insecure=true 
-            publish
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: >
+          ./gradlew
+          -Dorg.gradle.internal.publish.checksums.insecure=true
+          publish
   docs:
     if: github.event_name == 'push'
     needs: publish
@@ -61,18 +70,23 @@ jobs:
       GIT_USER_NAME: puneetbehl
       GIT_USER_EMAIL: behlp@unityfoundation.io
     steps:
-      - name: Checkout repository
+      - name: "📥 Checkout repository"
         uses: actions/checkout@v4
-      - name: Set up JDK
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: 17
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       - name: Generate Groovydoc
         id: groovydoc
         uses: gradle/gradle-build-action@v3
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         with:
           arguments: groovydoc
       - name: Publish to Github Pages
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,40 +12,43 @@ jobs:
       release_version: ${{ steps.release_version.outputs.value }}
       target_branch: ${{ steps.extract_branch.outputs.value }}
     env:
-      GIT_USER_NAME: puneetbehl
-      GIT_USER_EMAIL: behlp@unityfoundation.io
+      GIT_USER_NAME: 'grails-build'
+      GIT_USER_EMAIL: 'grails-build@users.noreply.github.com'
     steps:
-      - name: Checkout repository
+      - name: "📥 Checkout repository"
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.GH_TOKEN }}
-      - name: Set up JDK
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: 17
-      - name: Extract Target Branch
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+      - name: "📝 Store the target branch"
         id: extract_branch
         run: |
           echo "Determining Target Branch"
-          TARGET_BRANCH=`cat $GITHUB_EVENT_PATH | jq '.release.target_commitish' | sed -e 's/^"\(.*\)"$/\1/g'`
+          TARGET_BRANCH=${GITHUB_REF#refs/heads/}
           echo $TARGET_BRANCH
           echo "value=${TARGET_BRANCH}" >> $GITHUB_OUTPUT
-      - name: Set the current release version
+      - name: "📝 Store the current release version"
         id: release_version
         run: echo "value=${GITHUB_REF:11}" >> $GITHUB_OUTPUT
-      - name: Run pre-release
+      - name: "⚙️ Run pre-release"
         uses: micronaut-projects/github-actions/pre-release@master
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Run Assemble
+      - name: "🧩 Run Assemble"
         if: success()
         id: assemble
-        uses: gradle/gradle-build-action@v3
-        with:
-          arguments: assemble
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: ./gradlew assemble
       - name: Upload Distribution
         if: success()
         uses: actions/upload-artifact@v4
@@ -67,7 +70,8 @@ jobs:
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
           SIGNING_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
           SECRING_FILE: ${{ secrets.SECRING_FILE }}
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         with:
           arguments: |
             -Psigning.secretKeyRingFile=${{ github.workspace }}/secring.gpg 
@@ -93,7 +97,8 @@ jobs:
       - name: Nexus Staging Close And Release
         uses: gradle/gradle-build-action@v3
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_STAGING_PROFILE_ID: ${{ secrets.SONATYPE_STAGING_PROFILE_ID }}
@@ -114,8 +119,8 @@ jobs:
     permissions:
       contents: write
     env:
-      GIT_USER_NAME: puneetbehl
-      GIT_USER_EMAIL: behlp@unityfoundation.io
+      GIT_USER_NAME: 'grails-build'
+      GIT_USER_EMAIL: 'grails-build@users.noreply.github.com'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -131,7 +136,8 @@ jobs:
         id: groovydoc
         uses: gradle/gradle-build-action@v3
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         with:
           arguments: groovydoc
       - name: Publish to Github Pages
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,19 +49,18 @@ jobs:
           DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
           DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         run: ./gradlew assemble
-      - name: Upload Distribution
+      - name: "🚀 Upload Distribution"
         if: success()
         uses: actions/upload-artifact@v4
         with:
           name: grails-${{ steps.release_version.outputs.value }}.zip
           path: build/distributions/grails-${{ steps.release_version.outputs.value }}.zip
-      - name: Generate secring file
+      - name: "📝 Generate secring file"
         env:
           SECRING_FILE: ${{ secrets.SECRING_FILE }}
         run: echo $SECRING_FILE | base64 -d > ${{ github.workspace }}/secring.gpg
-      - name: Publish to Sonatype OSSRH
+      - name: "🚀 Publish to Sonatype OSSRH"
         id: publish
-        uses: gradle/gradle-build-action@v3
         env:
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
@@ -72,8 +71,8 @@ jobs:
           SECRING_FILE: ${{ secrets.SECRING_FILE }}
           DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
           DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
-        with:
-          arguments: |
+        run: |
+            ./gradlew
             -Psigning.secretKeyRingFile=${{ github.workspace }}/secring.gpg 
             publishToSonatype 
             closeSonatypeStagingRepository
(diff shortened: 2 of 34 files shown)


.github/workflows/gradle.yml
```diff
@@ -16,63 +16,77 @@ jobs:
...
-      - uses: actions/checkout@v4
-      - name: Set up JDK
+      - name: "📥 Checkout repository"
+        uses: actions/checkout@v4
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: ${{ matrix.java }}
```
matrei, line 23: Should we add 21 to the matrix? No need to have a matrix with one value.

.github/workflows/gradle.yml
```diff
@@ -16,63 +16,77 @@ jobs:
...
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: 17
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       - name: Generate Groovydoc
```
matrei: Use some nice emoji in step name :-)

.github/workflows/gradle.yml
```diff
@@ -16,63 +16,77 @@ jobs:
...
         id: groovydoc
         uses: gradle/gradle-build-action@v3
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         with:
```
matrei: Use `run: ./gradlew groovydoc`

.github/workflows/gradle.yml
```diff
@@ -16,63 +16,77 @@ jobs:
...
-          distribution: temurin
+          distribution: liberica
           java-version: 17
+      - name: "🐘 Setup Gradle"
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
       - name: Generate Groovydoc
         id: groovydoc
         uses: gradle/gradle-build-action@v3
```
matrei: No need for `uses:` here. Gradle is already setup.

.github/workflows/gradle.yml
```diff
@@ -16,63 +16,77 @@ jobs:
...
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
         with:
           arguments: groovydoc
       - name: Publish to Github Pages
```
matrei: Use some nice emoji in step name :-)

.github/workflows/gradle.yml
```diff
@@ -88,4 +102,4 @@ jobs:
           DOC_FOLDER: gh-pages
           COMMIT_EMAIL: behlp@unityfoundation.io
```
matrei: Change user

.github/workflows/gradle.yml
```diff
@@ -88,4 +102,4 @@ jobs:
...
           COMMIT_NAME: Puneet Behl
-          VERSION: ${{ needs.publish.outputs.release_version }}
\ No newline at end of file
+          VERSION: ${{ needs.publish.outputs.release_version }}
```
matrei: Is the `release_version` output defined in the `publish` step?

.github/workflows/release.yml
```diff
@@ -12,40 +12,43 @@ jobs:
...
         id: assemble
-        uses: gradle/gradle-build-action@v3
-        with:
-          arguments: assemble
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: ./gradlew assemble
       - name: Upload Distribution
```
matrei: Use nice emoji in all step names?

.github/workflows/release.yml
matrei, line 1: - `temurin` is still used in this file
- All jobs (`publish`, `release` and `docs`) still use old `gradle/gradle-build-action@v3`
- Use nice emoji in all step names.
- `SKIP_LATEST` param to `grails/github-pages-deploy-action` at the bottom has the wrong value.

.github/workflows/release.yml
```diff
@@ -114,8 +119,8 @@ jobs:
     permissions:
       contents: write
     env:
-      GIT_USER_NAME: puneetbehl
-      GIT_USER_EMAIL: behlp@unityfoundation.io
+      GIT_USER_NAME: 'grails-build'
```
matrei, line 125: Use this in the last step

README.md
```diff
@@ -41,6 +37,58 @@ grails-profile
...
+        websiteUrl = 'http://foo.com/myplugin'
+        license {
+            name = 'Apache-2.0'
+        }
+        issueTrackerUrl = 'http://github.com/myname/myplugin/issues'
+        vcsUrl = 'http://github.com/myname/myplugin'
+        title = "My plugin title"
+        desc = "My plugin description"
+        developers = [johndoe:"John Doe"]
+    }
```
matrei: - Use `https`
- Mix of single and double quotes for strings

README.md
```diff
@@ -41,6 +37,58 @@ grails-profile
...
+
+    grailsPublish {
+        githubSlug = 'foo/bar'
+        license {
+            name = 'Apache-2.0'
+        }
+        title = "My plugin title"
+        desc = "My plugin description"
+        developers = [johndoe:"John Doe"]
+    }
```
matrei: Mix of single and double quotes for string

settings.gradle
```diff
@@ -1,27 +1,34 @@
...
-//            taskInputFiles = true
-//        }
-//    }
-//}
+def isCI = System.getenv('CI') != null
+def isLocal = !isCI
+def isAuthenticated = System.getenv('DEVELOCITY_ACCESS_KEY') != null
+def isBuildCacheAuthenticated =
+		System.getenv('DEVELOCITY_BUILD_CACHE_NODE_USER') != null &&
+				System.getenv('DEVELOCITY_BUILD_CACHE_NODE_KEY') != null
```
matrei: Can we retract this line to align with the previous line? (I know IDEA wants to format it like this)

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,23 +36,17 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
 grailsPublish {
-    user = 'user'
-    key = 'key'
-    userOrg = 'my-company' // optional, otherwise published to personal bintray account
-    repo = 'plugins' // optional, defaults to 'plugins'
-
-
     websiteUrl = 'http://foo.com/myplugin'
     license {
         name = 'Apache-2.0'
```
matrei, line 53: - Use `https`
- Mix of single and double quotes

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -68,8 +61,6 @@ grailsPublish {
 or
 
 grailsPublish {
-    user = 'user'
-    key = 'key'
     githubSlug = 'foo/bar'
     license {
         name = 'Apache-2.0'
```
matrei, line 67: Mix of single and double quotes

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -97,41 +93,43 @@ BINTRAY_KEY=key
...
+        final String artifactoryUsername = project.findProperty("artifactoryPublishUsername") ?: System.getenv("ARTIFACTORY_USERNAME") ?: ''
+        final String artifactoryPassword = project.findProperty("artifactoryPublishPassword") ?: System.getenv("ARTIFACTORY_PASSWORD") ?: ''
+        final String artifactoryPublishUrl = project.findProperty("artifactoryPublishUrl") ?: System.getenv("ARTIFACTORY_URL") ?: ''
+        final String sonatypeNexusUrl = project.findProperty("sonatypeNexusUrl") ?: System.getenv("SONATYPE_NEXUS_URL") ?: ''
+        final String sonatypeSnapshotUrl = project.findProperty("sonatypeSnapshotUrl") ?: System.getenv("SONATYPE_SNAPSHOT_URL") ?: ''
+        final String sonatypeUsername = project.findProperty("sonatypeUsername") ?: System.getenv("SONATYPE_USERNAME") ?: ''
+        final String sonatypePassword = project.findProperty("sonatypePassword") ?: System.getenv("SONATYPE_PASSWORD") ?: ''
+        final String sonatypeStagingProfileId = project.findProperty("sonatypeStagingProfileId") ?: System.getenv("SONATYPE_STAGING_PROFILE_ID") ?: ''
```
matrei: Mix of single and double quotes

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
```diff
@@ -0,0 +1,6 @@
+package org.grails.gradle.plugin.publishing
+
+enum RepositoryType {
```
matrei: Are these really different types? Would `RepositoryTarget` be more accurate?

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
```diff
@@ -0,0 +1,6 @@
...
+    ARTIFACTORY, // standard maven repository such as artifactory
```
matrei: You say, "standard maven repository such as artifactory". But the enum is named `ARTIFACTORY`. And the environment variables are named `ARTIFACTORY_USERNAME` etc. Should we be more general in the naming if any "standard maven repository" can be used?

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
```diff
@@ -0,0 +1,6 @@
...
+    CENTRAL // maven central
```
matrei: Would `MAVEN_CENTRAL` be more descriptive? Isn't Sonatype Nexus more accurate? Lots of questions here, but I think it is crucial to get the naming right to avoid confusion.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -97,41 +93,43 @@ BINTRAY_KEY=key
...
                             credentials {
                                 username = artifactoryUsername
                                 password = artifactoryPassword
                             }
-                            url gpe.snapshotUrl? gpe.snapshotUrl : getDefaultGrailsCentralSnapshotRepo()
+
+                            if(!artifactoryPublishUrl) {
+                                throw new RuntimeException("Could not locate a project property of `artifactoryPublishUrl` or an environment variable of `ARTIFACTORY_URL` for the snapshot url")
+                            }
+                            url artifactoryPublishUrl
```
matrei: url is a setter, use =

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -97,41 +93,43 @@ BINTRAY_KEY=key
...
+                    System.setProperty('org.gradle.internal.publish.checksums.insecure', true as String)
                     repositories {
                         maven {
                             credentials {
                                 username = artifactoryUsername
                                 password = artifactoryPassword
                             }
-                            url gpe.snapshotUrl? gpe.snapshotUrl : getDefaultGrailsCentralSnapshotRepo()
+
+                            if(!artifactoryPublishUrl) {
```
matrei: Space after if

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -264,37 +262,35 @@ BINTRAY_KEY=key
                 }
             }
 
-            if (isRelease) {
-                pluginManager.apply(NexusPublishPlugin.class)
-                pluginManager.apply(SigningPlugin.class)
+            if (isRelease || (isSnapshot && snapshotType == RepositoryType.CENTRAL)) {
+                rootProjectPluginManager.apply(NexusPublishPlugin.class)
+                projectPluginManager.apply(SigningPlugin.class)
```
matrei: No need to use `.class` suffix

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -264,37 +262,35 @@ BINTRAY_KEY=key
...
 
                 extensionContainer.configure(SigningExtension, {
                     it.required = isRelease
                     it.sign project.publishing.publications.maven
                 })
 
-                project.tasks.withType(io.github.gradlenexus.publishplugin.InitializeNexusStagingRepository).configureEach {
-                    shouldRunAfter(project.tasks.withType(Sign))
+                project.rootProject.tasks.withType(io.github.gradlenexus.publishplugin.InitializeNexusStagingRepository).configureEach { io.github.gradlenexus.publishplugin.InitializeNexusStagingRepository task ->
```
matrei: Import `InitializeNexusStagingRepository`

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -264,37 +262,35 @@ BINTRAY_KEY=key
...
+                    task.setShouldRunAfter(project.tasks.withType(Sign))
```
matrei: `setShouldRunAfter(Iterable<?>)` is a setter, use `shouldRunAfter =`

matrei (commented):
Nice work James! This should hopefully make it easier to publish in the future.

It is hard to test these kind of changes.
Also, I think some projects may depend on it so it will be exciting to see if it is backwards compatible. We'll see 😄.
(I see now that you already mentioned this in your PR description 👍)

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
jdaugherty: One uses the maven publish plugin with no signing, the other uses the nexus publish plugin which is specific to maven central / sonatype.  I agree we should rename these though.

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
jdaugherty: I'm going to rename these "MAVEN_PUBLISH" and "NEXUS_PUBLISH" ... that way it aligns with both the repository target & it matches the gradle plugins so there isn't confusion

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryType.groovy
jdaugherty: I renamed these to: 

        MAVEN_PUBLISH,
        NEXUS_PUBLISH

.github/workflows/gradle.yml
jdaugherty: I don't see, but looking at the changes before this PR, i don't see it there either.  I think it's a mistake?  grails-core uses a third party action to publish in it's gradle.yml.  grails-gsp omits the version completely.  I intend to workthrough the github actions after the PR merge, so I'm going to remove this now unless you have an idea what this value should be.

jdaugherty:
@matrei I believe I've made all of the changes you requested.  

In terms of backwards compatibility, this will intentionally break the other projects initially - I moved the hardcoded grails repo URL out of the plugin so that it can be used externally.  If we define that URL at the organization level, I believe it may "just work", but when I merge this I assumed I'll go through each repo rapidly fixing or converting each repo to a simplified publish.  Thank you so much for your feedback!

.github/workflows/gradle.yml
```diff
@@ -12,70 +12,82 @@ jobs:
...
         id: groovydoc
-        uses: gradle/gradle-build-action@v3
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
-        with:
-          arguments: groovydoc
-      - name: Publish to Github Pages
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: ./gradle groovydoc
```
matrei: ./gradlew

.github/workflows/gradle.yml
```diff
@@ -12,70 +12,82 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: ['17']
+        java: ['17', '21']
     env:
       WORKSPACE: ${{ github.workspace }}
```
matrei: I don't think this is necessary, `github.workspace` should be the default anyway.

.github/workflows/gradle.yml
matrei: I checked what it does, and it looks like `VERSION` is only needed in `release.yml`. So it's OK to remove it here in `gradle.yml`.

.github/workflows/release.yml
```diff
@@ -67,9 +69,10 @@ jobs:
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
           SIGNING_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
           SECRING_FILE: ${{ secrets.SECRING_FILE }}
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
-        with:
-          arguments: |
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
+        run: |
```
matrei: You need to use `>` instead of `|` because this is one command, not multiple commands.

.github/workflows/release.yml
```diff
@@ -79,26 +82,31 @@ jobs:
...
         env:
-          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
+          DEVELOCITY_BUILD_CACHE_NODE_USER: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_USER }}
+          DEVELOCITY_BUILD_CACHE_NODE_KEY: ${{ secrets.GRADLE_ENTERPRISE_BUILD_CACHE_NODE_KEY }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_STAGING_PROFILE_ID: ${{ secrets.SONATYPE_STAGING_PROFILE_ID }}
-        with:
-          arguments: |
+        run: |
```
matrei: You need to use `>` instead of `|` because this is one command, not multiple commands.

.github/workflows/release.yml
```diff
@@ -114,38 +122,37 @@ jobs:
...
         with:
           token: ${{ secrets.GH_TOKEN }}
           ref: v${{ needs.publish.outputs.release_version }}
-      - name: Set up JDK
+      - name: "☕️ Setup JDK"
         uses: actions/setup-java@v4
         with:
-          distribution: temurin
+          distribution: liberica
           java-version: 17
```
matrei, line 137: You also need to do a step with `setup-gradle` here.

README.md
```diff
@@ -41,6 +37,61 @@ grails-profile
...
+        githubSlug = 'foo/bar'
+        license {
+            name = 'Apache-2.0'
+        }
+        title = 'My plugin title'
+        desc = 'My plugin description'
+        developers = [johndoe: 'John Doe']
+    }
+
+By default this plugin will publish to the specified `MAVEN_PUBLISH` instance for snapshots, and `NEXUS_PUBLISH` for releases.  To change the snapshot publish behavior, you can set the `snapshotRepoType` to `RepositoryTarget.NEXUS_PUBLISH`.
```
matrei: I see that you renamed to `RepositoryTarget`. Should the property also be renamed to `snapshotRepoTarget`?

settings.gradle
```diff
@@ -1,27 +1,34 @@
...
+		enabled = true
+		usernameAndPassword(
+				System.getenv('DEVELOCITY_BUILD_CACHE_NODE_USER') ?: '',
+				System.getenv('DEVELOCITY_BUILD_CACHE_NODE_KEY') ?: ''
+		)
+	}
+}
 
 
 rootProject.name = "grails-gradle-plugin"
```
matrei: Use single quotes?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishExtension.groovy
```diff
@@ -25,29 +25,23 @@ import org.gradle.util.ConfigureUtil
...
-     * The slug from github
-     */
-    String githubSlug
-
     /**
-     * The the publishing user
+     * Determines the snapshot repository to publish to
      */
-    String user
+    RepositoryTarget snapshotRepoType = RepositoryTarget.MAVEN_PUBLISH
```
matrei: Should this be named `snapshotRepoTarget`?

README.md
```diff
@@ -41,6 +37,61 @@ grails-profile
...
+    ARTIFACTORY_URL
+
+The `NEXUS_PUBLISH` Environment Variables are:
+
+    SONATYPE_NEXUS_URL
+    SONATYPE_SNAPSHOT_URL
+    SONATYPE_USERNAME
+    SONATYPE_PASSWORD
+    SONATYPE_STAGING_PROFILE_ID
+
```
matrei: I'm concerned about the clarity of our current repository terminology. We're using multiple terms that might confuse readers about their relationships:

Current Terms:
- `RepositoryTarget`
- `snapshotRepoType` (target vs type, they are all Maven (type) repos right? So we are targeting different repos)
- `MAVEN_PUBLISH`
- `NEXUS_PUBLISH`
- `ARTIFACTORY_*`
- `SONATYPE_*`

`MAVEN_PUBLISH` vs `NEXUS_PUBLISH` is counterintuitive (to me, if I was a new user), as I would use `NEXUS_PUBLISH` to publish to Maven Central.

Does it make sense with `RepositoryTarget.GENERIC` and `RepositoryTarget.NEXUS` with environment vars `PUBLISH_GENERIC_*` and `PUBLISH_NEXUS_*`?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,101 +37,100 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
 
         final ExtraPropertiesExtension extraPropertiesExtension = extensionContainer.findByType(ExtraPropertiesExtension)
 
-        extraPropertiesExtension.setProperty(SIGNING_KEY_ID, project.hasProperty(SIGNING_KEY_ID) ? project[SIGNING_KEY_ID] : System.getenv("SIGNING_KEY") ?: null)
-        extraPropertiesExtension.setProperty(SIGNING_PASSWORD, project.hasProperty(SIGNING_PASSWORD) ? project[SIGNING_PASSWORD] : System.getenv("SIGNING_PASSPHRASE") ?: null)
-        extraPropertiesExtension.setProperty(SIGNING_KEYRING, project.hasProperty(SIGNING_KEYRING) ? project[SIGNING_KEYRING] : System.getenv("SIGNING_KEYRING") ?: null)
-
+        extraPropertiesExtension.setProperty(SIGNING_KEY_ID, project.hasProperty(SIGNING_KEY_ID) ? project[SIGNING_KEY_ID] : System.getenv('SIGNING_KEY') ?: null)
+        extraPropertiesExtension.setProperty(SIGNING_PASSWORD, project.hasProperty(SIGNING_PASSWORD) ? project[SIGNING_PASSWORD] : System.getenv('SIGNING_PASSPHRASE') ?: null)
+        extraPropertiesExtension.setProperty(SIGNING_KEYRING, project.hasProperty(SIGNING_KEYRING) ? project[SIGNING_KEYRING] : System.getenv('SIGNING_KEYRING') ?: null)
```
matrei: Use:
```groovy
extraPropertiesExtension.setProperty(SIGNING_KEY_ID, project.findProperty(SIGNING_KEY_ID) ?: System.getenv('SIGNING_KEY'))
extraPropertiesExtension.setProperty(SIGNING_PASSWORD, project.findProperty(SIGNING_PASSWORD) ?: System.getenv('SIGNING_PASSPHRASE'))
extraPropertiesExtension.setProperty(SIGNING_KEYRING, project.findProperty(SIGNING_KEYRING) ?: System.getenv('SIGNING_KEYRING'))
```

It should do the same thing.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,101 +37,100 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
 
         project.afterEvaluate {
-            boolean isSnapshot = project.version.endsWith("SNAPSHOT")
+            RepositoryTarget snapshotType = gpe.snapshotRepoType
```
matrei: `snapshotTarget`?

src/main/groovy/org/grails/gradle/plugin/publishing/RepositoryTarget.groovy
```diff
@@ -0,0 +1,6 @@
+package org.grails.gradle.plugin.publishing
+
+enum RepositoryTarget {
+    MAVEN_PUBLISH,
+    NEXUS_PUBLISH
```
matrei: See comment in readme about naming

README.md
matrei: After giving this some more thought, I’d like to better understand the problem we’re solving here.
If I understand correctly, the goal is to publish snapshot and release artifacts to different repositories.

Would it make sense to name them `RepositoryTarget.SNAPSHOT` and `RepositoryTarget.RELEASE`?
Similarly, we could use environment variables like `PUBLISH_SNAPSHOT_*` and `PUBLISH_RELEASE_*`.

It’s getting late here, so there’s a chance I’m overlooking something obvious 😄. Let me know your thoughts!

matrei: We could make signing an attribute, `default = true` for release and `false` for snapshot.
Depending on the repo url or perhaps attribute (`isNexusRepo` or something), use the nexus publish plugin or not.

jdaugherty: couple of problems:

1. every other plugin will likely only publish to maven central for snapshots & releases
2. grails currently publishes to artifactory (repo.grails.org) for snapshots (valid maven repo)
3. eventually we may want to use maven central's snapshot location for grails; in some case we already did this (rest plugin)
4. plugins may even want to publish to github, which only requires maven publish

The setup for maven central is cumbersome - especially in multiproject builds due to its signing requirements and the close/stage steps.  There are a lot of legacy plugins that would benefit from an easier to configure setup like grailsPublish - we had to make a blog post because it was so complicated.

My goal was:
1. to make this plugin public since the majority of the logic already is here.  That then allows people to easily migrate older plugins to support publishing with minimal configuration.  
2. support multiproject builds
3. cleanup the legacy fields that no longer apply (they were left around when it was internalized)
4. have a way for us to easily switch snapshot publish locations (just change the type and the appropriate url from the org would be used)
5. This also would consolidate all of our publishing logic to one location, this plugin.

As part of the milestone we could announce this and a sample github action to make migrating easier.

jdaugherty: To your snapshot/ release target comments, that does not makes sense because its valid to publish snapshots to a maven repo or maven central's snapshot location.  We ultimately are configuring both the snapshot and release locations - which determines which plugin we use.

jdaugherty: At the heart of this plugin we are making it use different gradle plugins.  Maybe the RepositoryTarget should be based on that

jdaugherty: @matrei can you check this again?  See if this is better.

README.md
```diff
@@ -41,6 +37,58 @@ grails-profile
...
+        githubSlug = 'foo/bar'
+        license {
+            name = 'Apache-2.0'
+        }
+        title = 'My plugin title'
+        desc = 'My plugin description'
+        developers = [johndoe: 'John Doe']
+    }
+
+By default this plugin will publish to the specified `MAVEN_PUBLISH` instance for snapshots, and `NEXUS_PUBLISH` for releases.  To change the snapshot publish behavior, set `snapshotRepoType` to `PublishType.NEXUS_PUBLISH`. To change the release publish behavior,  set `releaseRepoType` to `PublishType.NEXUS_PUBLISH`.
```
matrei: The last sentence, should it not be: To change the release publish behavior,  set `releaseRepoType` to `PublishType.MAVEN_PUBLISH`

README.md
```diff
@@ -41,6 +37,58 @@ grails-profile
...
+    MAVEN_PUBLISH_URL
+
+`NEXUS_PUBLISH` Environment Variables are:
+
+    NEXUS_PUBLISH_URL
+    NEXUS_PUBLISH_SNAPSHOT_URL
+    NEXUS_PUBLISH_USERNAME
+    NEXUS_PUBLISH_PASSWORD
+    NEXUS_PUBLISH_STAGING_PROFILE_ID
+
```
matrei, line 91: Same ordering for both types?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,101 +36,109 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
-BINTRAY_USER=user
-BINTRAY_KEY=key
+`NEXUS_PUBLISH` Environment Variables are:
+    NEXUS_PUBLISH_URL
+    NEXUS_PUBLISH_SNAPSHOT_URL
+    NEXUS_PUBLISH_USERNAME
+    NEXUS_PUBLISH_PASSWORD
+    NEXUS_PUBLISH_STAGING_PROFILE_ID
 """
     }
```
matrei, line 90: Same ordering for both types?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -157,43 +164,40 @@ BINTRAY_KEY=key
...
                                 pomNode.dependencyManagement[0].replaceNode {}
                             }
 
                             if (gpe != null) {
                                 pomNode.children().last() + {
                                     def title = gpe.title ?: project.name
-                                    delegate.name title
-                                    delegate.description gpe.desc ?: title
+                                    delegate.name = title
+                                    delegate.description = gpe.desc ?: title
```
matrei: I don't think you can't use = on `delegate`. Same applies for every occurrence below. You want to trigger `methodMissing` and that will not happen if you call a property setter.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,101 +36,109 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
-            final PluginManager pluginManager = project.getPluginManager()
-            pluginManager.apply(MavenPublishPlugin.class)
 
-            // Remove "plain" archive classifier, unless bootJar or bootWar are enabled
-            if (!taskContainer.findByName("bootJar")?.enabled && !taskContainer.findByName("bootWar")?.enabled) {
-                (taskContainer.findByName("jar") as Jar).archiveClassifier.set("")
-            }
+            boolean mavenPublish = (isSnapshot && snapshotPublishType == PublishType.MAVEN_PUBLISH) || (isRelease && releasePublishType == PublishType.MAVEN_PUBLISH)
+            boolean sonatypePublish = (isSnapshot && snapshotPublishType == PublishType.NEXUS_PUBLISH) || (isRelease && releasePublishType == PublishType.NEXUS_PUBLISH)
+
```
matrei, line 122: nexusPublish?

matrei (changes requested):

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -37,101 +36,116 @@ import static com.bmuschko.gradle.nexus.NexusPlugin.getSIGNING_PASSWORD
...
+        boolean mavenPublish = (isSnapshot && snapshotPublishType == PublishType.MAVEN_PUBLISH) || (isRelease && releasePublishType == PublishType.MAVEN_PUBLISH)
+        boolean nexusPublish = (isSnapshot && snapshotPublishType == PublishType.NEXUS_PUBLISH) || (isRelease && releasePublishType == PublishType.NEXUS_PUBLISH)
+
+        final PluginManager projectPluginManager = project.getPluginManager()
+        final PluginManager rootProjectPluginManager = project.rootProject.getPluginManager()
+
+        // Required for the pom always
+        projectPluginManager.apply(MavenPublishPlugin)
 
+        if(nexusPublish) {
```
matrei: Space after if

matrei (approved):
You could also change the year in the license headers of changes files.

Let's try this baby out!

PR #367, opened 2024-12-08
Support overriding the groupId for the published artifact

jdaugherty:
Allows specifying the groupId for the published artifact


matrei (approved):
This can come in handy.
You don't want to add `artifactId` overriding while you are at it?

jdaugherty:
> This can come in handy. You don't want to add `artifactId` overriding while you are at it?

Done.  Thank you.

PR #368, opened 2024-12-08
Always register the integration scope & tasks

jdaugherty:


jdaugherty:
While convering to grails 7, I couldn't get a project to run.  I kept getting this error: 

        Could not find method integrationTestImplementation() for arguments [DefaultExternalModuleDependency{group='org.grails.plugins', name='geb', version='null', configuration='default'}] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

It was because the integration-test directory didn't exist so the integration plugin wasn't applied, and thus the scope didn't exist so gradle would fail because we added `integrationTestImplementation testFixtures("org.grails.plugins:geb")`

jamesfredley (approved):

PR #371, opened 2024-12-13
Add logging & improve version detection for plugin{} block usage

jdaugherty:


jdaugherty:
As part of the profile publishing we discovered that it's possible to leave a version unspecified before the plugin is applied. this change fixes that & adds logging for debugging

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
         extraPropertiesExtension.setProperty(SIGNING_PASSWORD, project.findProperty(SIGNING_PASSWORD) ?: System.getenv('SIGNING_PASSPHRASE'))
         extraPropertiesExtension.setProperty(SIGNING_KEYRING, project.findProperty(SIGNING_KEYRING) ?: System.getenv('SIGNING_KEYRING'))
 
-
         PublishType snapshotPublishType = gpe.snapshotPublishType
         PublishType releasePublishType = gpe.releasePublishType
 
-        boolean isSnapshot = project.version.endsWith('SNAPSHOT')
+        String detectedVersion = (project.version == Project.DEFAULT_VERSION ? (project.hasProperty('projectVersion') ? project.findProperty('projectVersion') : null) : project.version) as String
```
matrei: `(project.hasProperty('projectVersion') ? project.findProperty('projectVersion') : null)` can be replaced with `project.findProperty('projectVersion')`. It returns the value or `null` if it does not exist.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
+        if(detectedVersion == Project.DEFAULT_VERSION) {
+            throw new IllegalStateException("Project `${project.name}` has an unspecified version (neither version or property projectVersion is defined) so release publishing cannot be determined.")
+        }
+
+        boolean isSnapshot = detectedVersion.endsWith('SNAPSHOT')
+        if(isSnapshot) {
```
matrei: Space after if (more occurrences below)

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
+            project.rootProject.getLogger().info("Snapshot detected for `${project.name}`:${detectedVersion}")
```
matrei: Use groovy property syntax for `getLogger()`. Space after colon in log message. (all occurrences below)

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
         extraPropertiesExtension.setProperty(SIGNING_KEYRING, project.findProperty(SIGNING_KEYRING) ?: System.getenv('SIGNING_KEYRING'))
 
-
         PublishType snapshotPublishType = gpe.snapshotPublishType
         PublishType releasePublishType = gpe.releasePublishType
 
-        boolean isSnapshot = project.version.endsWith('SNAPSHOT')
+        String detectedVersion = (project.version == Project.DEFAULT_VERSION ? (project.hasProperty('projectVersion') ? project.findProperty('projectVersion') : null) : project.version) as String
+        if(detectedVersion == Project.DEFAULT_VERSION) {
+            throw new IllegalStateException("Project `${project.name}` has an unspecified version (neither version or property projectVersion is defined) so release publishing cannot be determined.")
```
matrei: No need to use curly braces for variable substitution in GString. (all occurrences below)

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
+        boolean isSnapshot = detectedVersion.endsWith('SNAPSHOT')
+        if(isSnapshot) {
+            project.rootProject.getLogger().info("Snapshot detected for `${project.name}`:${detectedVersion}")
+        }
         boolean isRelease = !isSnapshot
+        if(isRelease) {
+            project.rootProject.getLogger().info("Release detected for `${project.name}`:${detectedVersion}")
+        }
+
         boolean mavenPublish = (isSnapshot && snapshotPublishType == PublishType.MAVEN_PUBLISH) || (isRelease && releasePublishType == PublishType.MAVEN_PUBLISH)
```
matrei: Rename variable to `useMavenPublish`?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
+        if(mavenPublish) {
+            project.rootProject.getLogger().info("Maven Publish is enabled for `${project.name}` with version ${detectedVersion}")
+        }
         boolean nexusPublish = (isSnapshot && snapshotPublishType == PublishType.NEXUS_PUBLISH) || (isRelease && releasePublishType == PublishType.NEXUS_PUBLISH)
```
matrei: Rename variable to `useNexusPublish`?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,14 +113,31 @@ The credentials and connection url must be specified as a project property or an
...
+        if(nexusPublish) {
+            project.rootProject.getLogger().info("Nexus Publish is enabled for `${project.name}` with version ${detectedVersion}")
+        }
 
         final PluginManager projectPluginManager = project.getPluginManager()
         final PluginManager rootProjectPluginManager = project.rootProject.getPluginManager()
```
matrei: Use groovy property syntax for `getPluginManager()`

matrei (changes requested):
I think it is the `version` property that is actually used by `maven-publish` so if only `projectVersion` is set, we need to also set the `version` property.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
jdaugherty: There's actually a bug here, it needs to return the unspecified version

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
jdaugherty: I'm fine with that for single variables, but I highly prefer the brace syntax when accessing nested properties.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
-
         PublishType snapshotPublishType = gpe.snapshotPublishType
         PublishType releasePublishType = gpe.releasePublishType
 
-        boolean isSnapshot = project.version.endsWith('SNAPSHOT')
+        String detectedVersion = (project.version == Project.DEFAULT_VERSION ? (project.findProperty('projectVersion') ?: Project.DEFAULT_VERSION) : project.version) as String
+        if (detectedVersion == Project.DEFAULT_VERSION) {
+            throw new IllegalStateException("Project `${project.name}` has an unspecified version (neither `version` or the property `projectVersion` is defined). Release state cannot be determined.")
+        }
+        project.rootProject.logger.info("Detected Version $detectedVersion for Project `${project.name}`")
```
matrei: Version $detectedVersion detected for project `${project.name}`

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
+
+        boolean isSnapshot = detectedVersion.endsWith('SNAPSHOT')
+        if (isSnapshot) {
+            project.rootProject.logger.info("Snapshot detected for Project `${project.name}`")
```
matrei: Snapshot version detected for project `${project.name}`

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
-        boolean nexusPublish = (isSnapshot && snapshotPublishType == PublishType.NEXUS_PUBLISH) || (isRelease && releasePublishType == PublishType.NEXUS_PUBLISH)
+        if (isRelease) {
+            project.rootProject.logger.info("Release detected for Project `${project.name}`")
+        }
 
-        final PluginManager projectPluginManager = project.getPluginManager()
-        final PluginManager rootProjectPluginManager = project.rootProject.getPluginManager()
+        boolean useMavenPublish = (isSnapshot && snapshotPublishType == PublishType.MAVEN_PUBLISH) || (isRelease && releasePublishType == PublishType.MAVEN_PUBLISH)
+        if (useMavenPublish) {
+            project.rootProject.logger.info("Maven Publish is enabled for Project `${project.name}`")
```
matrei: Maven Publish is enabled for project `${project.name}`

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
+        }
+        boolean useNexusPublish = (isSnapshot && snapshotPublishType == PublishType.NEXUS_PUBLISH) || (isRelease && releasePublishType == PublishType.NEXUS_PUBLISH)
+        if (useNexusPublish) {
+            project.rootProject.logger.info("Nexus Publish is enabled for Project `${project.name}`")
```
matrei: Nexus Publish is enabled for project `${project.name}`

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
+        final PluginManager projectPluginManager = project.pluginManager
         projectPluginManager.apply(MavenPublishPlugin)
 
-        if (nexusPublish) {
+        if (useNexusPublish) {
+            // The nexus plugin is special since it must always be applied to the root project.
+            // Handle when multiple subprojects exist and grailsPublish is defined in each one instead of at the root.
+            final PluginManager rootProjectPluginManager = project.rootProject.pluginManager
+            boolean hasNexusPublishApplied = rootProjectPluginManager.hasPlugin(NEXUS_PUBLISH_PLUGIN_ID)
+            if (hasNexusPublishApplied) {
```
matrei, line 159: Nice! This fixes the problem in the `grails-geb` test project, right?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -111,22 +120,46 @@ The credentials and connection url must be specified as a project property or an
...
+                project.rootProject.logger.info("Nexus Publish Plugin already applied to root project")
+            }
+
             rootProjectPluginManager.apply(NexusPublishPlugin)
```
matrei: But must this not be in an else clause?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -167,8 +202,9 @@ The credentials and connection url must be specified as a project property or an
                             }
 
                             if (!mavenPublishUrl) {
-                           //     throw new RuntimeException('Could not locate a project property of `mavenPublishUrl` or an environment variable of `MAVEN_PUBLISH_URL`')
+                                throw new RuntimeException('Could not locate a project property of `mavenPublishUrl` or an environment variable of `MAVEN_PUBLISH_URL`.  A URL is required for maven publishing.')
```
matrei: Double space

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
jdaugherty: No, it fixes another problem with one root and many nested subprojects.  I am still trying to figure out that one

matrei:
> I think it is the `version` property that is actually used by `maven-publish` so if only `projectVersion` is set, we need to also set the `version` property.

@jdaugherty I think this is still an issue. What happens if only `projectVersion` is set? `detectedVersion` become `projectVersion`. But `version` will still be `unspecified`. Isn't `version` the property that is used by `maven-publish`?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
jdaugherty: I tried that & I was hoping it would fix the Geb project, but it doesn't.  

This actually works when you have a project structure so that there is a common root build.gradle and several subprojects with build.gradle in them.  You can reapply it without issue in my testing.  It wouldn't hurt to have it an else, so I'm going to add it there.

jdaugherty:
> > I think it is the `version` property that is actually used by `maven-publish` so if only `projectVersion` is set, we need to also set the `version` property.
> 
> @jdaugherty I think this is still an issue. What happens if only `projectVersion` is set? `detectedVersion` become `projectVersion`. But `version` will still be `unspecified`. Isn't `version` the property that is used by `maven-publish`?

The purpose of this plugin is it needs to determine if it's a release or a snapshot so it can handle the nexus publishing vs snapshot logic.  It currently does that based on the project.version property.  The issue is if you apply this plugin in the rootProject under the plugins {} block, it will be evaluated before project.version can be set.  Worse, we can't do this check in the afterEvaluate because at that point it's too late to apply the nexus plugin.  

I added the projectVersion check as a work around since we typically use that variable.  

I think the real solution is to move the release state to a variable and then we wouldn't need the version.  Basically, we could do those checks outside of gradle.  I was hesitant to make this change for this release since it means every action has to change again.   Although, I think realistically we always know based on the action if it's a release or snapshot so I really like this idea instead of relying on the version #.

matrei (approved):

PR #372, opened 2024-12-13
Comment out validation for now

jdaugherty:
Commenting out the validation for now.


jamesfredley (approved):

jamesfredley:
Could this be a warning instead of an exception?

jdaugherty:
It needs to be an exception, but i need to rework this to be a task dependency instead of doing it at the time of configuration.  The only reason I'm trying to throw the exception is to give a better error.  It will still error with this commented out.

PR #375, opened 2024-12-19
End to End testing for the publish plugin & clean-up of previously failed changes

### LINKED ISSUE #3: Grails Publish Improvements
There are 2 bugs I'm tracking down:
1. Sometimes when using the newer `plugins {}` method instead of `apply plugin: 'org.grails.grails-publish'` in a multiproject build it fails.
2. Having a simple build file at the root and including the plugin under a subproject can cause a failure (grails-spring-security-acl). For now the fix is to include the plugin at the root dependency and apply in the subproject.
I'll create a ticket to look at these before the 7.0.0 release.

### LINKED ISSUE #364: Grails Publish Improvements
There are 2 bugs I'm tracking down:
1. Sometimes when using the newer `plugins {}` method instead of `apply plugin: 'org.grails.grails-publish'` in a multiproject build it fails.
2. Having a simple build file at the root and including the plugin under a subproject can cause a failure (grails-spring-security-acl). For now the fix is to include the plugin at the root dependency and apply in the subproject.
I'll create a ticket to look at these before the 7.0.0 release.

jdaugherty:
This change encompasses: 
* not running publish action on forks
* not running the release notes action on fork (since secrets aren't available in forks / pull requests)
* end to end testing support for any gradle plugin in this project (including logging for tests in the action)
* added several tests for the publish plugin based on past issues
* publishes of a release using MAVEN_PUBLISH should sign the build
* added back the url validation when using maven publish (previously validation was happening at configure time which caused failures in assemble)
* add a warning when projectVersion is being used instead of project.version (longer term I think we may just want to make the release/snapshot determination an environment variable / system property instead of hard coded logic)
* updated some log statements to match the style of previous code review comments

This ticket does not fully fix apache/incubator-grails-gradle-publish#3 but it does help us diagnosis issues as they arise.  There are 4 tests marked as pending feature which documents the issues found in apache/incubator-grails-gradle-publish#3.  I think it may be reasonable that this can't be fixed, but I intend to investigate these further now that we have tests.


jdaugherty:
This now supports build release/snapshot detection via environment variable too.

jamesfredley (approved):

matrei (approved):

PR #380, opened 2024-12-23
Exclude `logback-spring.xml` in plugin jars

jdaugherty:


jdaugherty:
@matrei  since the docs are going to reference logback-spring.xml and our examples already do, I'm going ahead and proposing this separate from your other PR to clean up the jar process.

matrei (approved):

PR #383, opened 2024-12-24
Fix profile publish plugin

jdaugherty:


jdaugherty:
When working on spring security rest's profile publish: 

      No signature of method: org.grails.gradle.plugin.profiles.GrailsProfilePublishGradlePlugin.validateProjectState() is applicable for argument types: (org.gradle.api.internal.project.DefaultProject_Decorated) values: [project ':spring-security-rest-testapp-profile']

jdaugherty:
Merging this so I can get spring-security-rest fixed.  @jamesfredley I think this may be another groovy issue.  The methods that are private should be found on the parent class since that method is what's invoked.

PR #388, opened 2025-01-10
Fix `grails-plugin.xml` generation and `importJavaTime` setting

matrei:
Related to https://github.com/grails/grails-core/issues/13959

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,6 +75,8 @@ java {
 }
 
 dependencies {
+    implementation platform("org.grails:grails-bom:$grailsVersion")
+
     // compile grails-gradle-plugin with the Groovy version provided by Gradle
     // to ensure build compatibility with Gradle, currently Groovy 3.0.x
     // see: https://docs.gradle.org/current/userguide/compatibility.html#groovy
--- a/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/core/GrailsExtension.groovy
@@ -60,7 +60,7 @@ class GrailsExtension {
     /**
      * Whether java.time.* package should be a default import package
      */
-    boolean importJavaTime = true
+    boolean importJavaTime = false
 
     /**
      * Configure the reloading agent
--- a/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/core/GrailsGradlePlugin.groovy
@@ -289,24 +289,29 @@ class GrailsGradlePlugin extends GroovyPlugin {
                 }
             }
         }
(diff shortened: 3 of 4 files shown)


jdaugherty (approved):

codeconsole:
@matrei 
What exactly is this fixing and how is it fixing it?
Why default to false?  What is the conflict?

matrei:
@codeconsole Sorry, it is described in the commit messages.

jdaugherty (approved):

codeconsole:
@matrei but why default to false now? appending doesn't resolve the issue?

codeconsole:
https://github.com/grails/grails-gradle-plugin/pull/346

jdaugherty:
We should be defaulting these to false to match the previous behavior.  Especially since the behavior has caused an issue in this area.

codeconsole:
@jdaugherty it really has nothing to do with what it does but rather how it did it which seems would no longer be an issue.

@matrei I wonder if we should use the same naming convention for the file?
```groovy
def configFile = project.file("$project.buildDir/config.groovy")
```

https://github.com/grails/grails-gradle-plugin/blob/e62f68b46bc3bd8051121c7e65476c22022308d4/src/main/groovy/org/grails/gradle/plugin/core/GrailsPluginGradlePlugin.groovy#L259-L286

jamesfredley:
In addition to fixing `grails-plugin.xml` generation, we need to make sure this plays nicely with `joda-time` which also has a `LocalDate` class.  Many projects moved to `joda-time`, before `java.time` was released and before their projects were able to update to Java 8 and some chose to stick with `joda-time`, since it was widely spread across their codebase and not worth switching to `java.time`

https://www.joda.org/joda-time/apidocs/org/joda/time/LocalDate.html

If `boolean importJavaTime = false` is required to fix `grails-plugin.xml` generation it likely applies to other situations and we do not want to break those situations just to save adding `import java.time.LocalDate` in classes.

codeconsole (changes requested):
I am fine with changing default behavior back to disabled. It is just a 1 line property to turn it on, but I think it might be a good idea to use the same naming convention of the build script to match what is used for plugins.

matrei:
> I think it might be a good idea to use the same naming convention of the build script to match what is used for plugins.

@codeconsole Can you describe more in detail? I don't understand how you mean.

codeconsole:
> > I think it might be a good idea to use the same naming convention of the build script to match what is used for plugins.
> 
> @codeconsole Can you describe more in detail? I don't understand how you mean.

My mistake, you are actually already doing what I am suggesting... I was in the middle of something else when I first reviewed this and overlooked that part of the code.

Looks good

codeconsole (approved):

jeffscottbrown:
> I am fine with changing default behavior back to disabled.

+1

PR #394, opened 2025-01-24
Source published dependency versions from resolved artifacts in published POMs

matrei:
Updated publishing to retrieve published dependency versions directly from resolved artifacts, when generating POMs, eliminating reliance on the Spring Dependency Management plugin. This ensures greater independence and flexibility in managing versions, as the publishing plugin works without the Spring Dependency Management plugin being applied.

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2024 original authors
+ * Copyright 2015-2025 original authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jdaugherty:
Isn't the spring dependency plugin still going to be applied as part of applying the boot plugin?  This PR seems to indicate it eliminates that requirement, but I don't see any changes preventing the application of that in the gradle core plugin?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -344,13 +345,34 @@ Note: if project properties are used, the properties must be defined prior to ap
...
+                                def groupId = ((dependencyNode as Node).getAt(groupIdQName).first() as Node).text()
+                                def artifactId = ((dependencyNode as Node).getAt(artifactIdQName).first() as Node).text()
+                                def resolvedArtifacts = project.configurations.compileClasspath.resolvedConfiguration.resolvedArtifacts +
+                                                        project.configurations.runtimeClasspath.resolvedConfiguration.resolvedArtifacts
+                                def managedVersion = resolvedArtifacts.find {
+                                    it.moduleVersion.id.group == groupId &&
+                                    it.moduleVersion.id.name == artifactId
+                                }?.moduleVersion?.id?.version
+                                if (!managedVersion) {
+                                    throw new RuntimeException("No version found for dependency $groupId:$artifactId.")
```
jdaugherty, line 368: Won't gradle tell us this?  Why is this required in the bom world?  What am I missing here?

jdaugherty (approved):

PR #396, opened 2025-01-31
Include `testFixtures` configurations in POM version resolution

matrei:
Previously, only `compileClasspath` and `runtimeClasspath` were used to resolve dependencies. This caused issues with `grails-geb`, which relies on the `testFixtures` Gradle feature.

Now, dependencies defined in `testFixtures` are also included, ensuring proper resolution.

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
@@ -360,6 +360,10 @@ Note: if project properties are used, the properties must be defined prior to ap
                                 def artifactId = ((dependencyNode as Node).getAt(artifactIdQName).first() as Node).text()
                                 def resolvedArtifacts = project.configurations.compileClasspath.resolvedConfiguration.resolvedArtifacts +
                                                         project.configurations.runtimeClasspath.resolvedConfiguration.resolvedArtifacts
+                                if (project.configurations.hasProperty('testFixturesCompileClasspath')) {
+                                    resolvedArtifacts += project.configurations.testFixturesCompileClasspath.resolvedConfiguration.resolvedArtifacts +
+                                                         project.configurations.testFixturesRuntimeClasspath.resolvedConfiguration.resolvedArtifacts
+                                }
                                 def managedVersion = resolvedArtifacts.find {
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -360,6 +360,10 @@ Note: if project properties are used, the properties must be defined prior to ap
                                 def artifactId = ((dependencyNode as Node).getAt(artifactIdQName).first() as Node).text()
                                 def resolvedArtifacts = project.configurations.compileClasspath.resolvedConfiguration.resolvedArtifacts +
                                                         project.configurations.runtimeClasspath.resolvedConfiguration.resolvedArtifacts
+                                if (project.configurations.hasProperty('testFixturesCompileClasspath')) {
```
jdaugherty, line 363: After seeing @codeconsole's comments, I'm still confused why we need this.  This code looks like it's collapsing all of the classpaths together.  Wouldn't it be better to fix gradle so that a bom is set and those versions are pulled from the bom?  What am I missing on why we need this code?

matrei: This code is specifically for POM generation in the grails-publish plugin. Without it, dependencies that inherit their versions from a BOM would be missing a `<version>` element in the generated POM, leading to issues like:
```xml
<dependency>
  <groupId>org.gebish</groupId>
  <artifactId>geb-spock</artifactId>
  <scope>compile</scope>
  <optional>true</optional>
</dependency>
```

This logic identifies such dependencies and explicitly adds the resolved version from the BOM:
```xml
<dependency>
  <groupId>org.gebish</groupId>
  <artifactId>geb-spock</artifactId>
  <scope>compile</scope>
  <optional>true</optional>
  <version>7.0</version>
</dependency>
```

Previously (before #394), the Spring Dependency Management Plugin was used to do the same thing, but it’s not used consistently across all Grails libraries (for example `data-mapping`). Since some libraries don’t apply the Grails Gradle Plugin—and thus don’t get dependency management—their generated POMs were broken due to missing version elements.

This code ensures correct POM generation even when the Spring Dependency Management Plugin is not applied.

jdaugherty (approved):

PR #398, opened 2025-02-11
Fix ast sources in javadoc & source jars

jdaugherty:
I couldn't figure out why the AST files weren't being included in the source or javadoc jars.  Turns out it's because we're searching for the tasks immediately, so if they weren't defined prior to the gradle plugin being applied, they wouldn't get included.  This PR changes those tasks to be run after evaluation, which means you can now use composition in your gradle files and the task will be found correctly.

This PR also fixes the publish plugin so that it doesn't try to add the task artifacts - when you invoke `publication.from project.components.java` that will add the source jars, javadoc, and class jars.  There's no reason to add them again.  This actually causes withSourceJars() & withJavadocJar() to fail if defined.

This PR also moves away from toolchain so the matrix builds work correctly & use the right version of java.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,11 +65,8 @@ if (isReleaseVersion) {
     apply plugin: 'signing'
 }
 
+compileJava.options.release = 17
 java {
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(17)
-    }
-
     withJavadocJar()
     withSourcesJar()
 }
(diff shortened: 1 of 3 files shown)


jdaugherty:
This will fix the grails-data-mapping project so that we can add back: 

     java {
         withSourcesJar()
         withJavadocJar()
     }

matrei (approved):
It's really hard to review these kind of changes. I'll have to try it out. If you have tested the output and think everything works as it should, I'll approve. If you want me to do more thorough testing, I'll have to defer until tomorrow.

jdaugherty:
> It's really hard to review these kind of changes. I'll have to try it out. If you have tested the output and think everything works as it should, I'll approve. If you want me to do more thorough testing, I'll have to defer until tomorrow.

I used the grails-cache project to test this.  I also extracted the source & javadoc jars to ensure they're including everything now.  Let's merge and circle back if there are any issues found.

PR #399, opened 2025-02-12
Improve the publishing plugin to ensure sources & javadoc are published by default

jdaugherty:
@matrei we need to start switching to the publish plugin everywhere after we merge this.  It will ensure we're correctly publishing source & javadoc for all projects if we do.

### DIFF
--- a/src/e2eTest/resources/publish-projects/other-artifacts/explicit-jar-creation-without-gradle-assistance/build.gradle
+++ b/src/e2eTest/resources/publish-projects/other-artifacts/explicit-jar-creation-without-gradle-assistance/build.gradle
@@ -1,0 +1,53 @@
+buildscript {
+    repositories {
+        maven { url "${System.getenv('LOCAL_MAVEN_PATH')}" }
+        maven { url = 'https://repo.grails.org/grails/core' }
+    }
+    dependencies {
+        classpath "org.grails:grails-gradle-plugin:$grailsGradlePluginVersion"
+    }
+}
+
+allprojects {
+    repositories {
+        maven { url = 'https://repo.grails.org/grails/core' }
+    }
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
+}
+
+apply plugin: 'org.grails.grails-publish'
+grailsPublish {
+    githubSlug = 'grails/grails-gradle-plugin'
+    license {
+        name = 'Apache-2.0'
+    }
+    title = 'Grails Gradle Plugin - Example Project'
+    desc = 'A testing project for the grails gradle plugin'
+    developers = [
+            jdaugherty: 'James Daugherty',
+    ]
+}
+
+// Add sources and javadoc jar tasks
+if (!tasks.findByName('sourcesJar')) {
+    tasks.register('sourcesJar', Jar) {
+        archiveClassifier = 'sources'
+        from sourceSets.main.allSource
+    }
+}
+if (!tasks.findByName('javadocJar')) {
+    tasks.register('javadocJar', Jar) {
+        dependsOn 'javadoc'
+        archiveClassifier = 'javadoc'
+        from javadoc.destinationDir
+    }
+}
--- a/src/e2eTest/resources/publish-projects/other-artifacts/explicit-jar-creation-without-gradle-assistance/build.gradle
+++ b/src/e2eTest/resources/publish-projects/other-artifacts/explicit-jar-creation-without-gradle-assistance/build.gradle
@@ -21,7 +21,7 @@ apply plugin: 'java-library'
 apply plugin: 'groovy'
 
 dependencies {
-    implementation 'org.apache.groovy:groovy-all:4.0.24'
+    implementation "org.apache.groovy:groovy-all:$groovyVersion"
 }
 
 apply plugin: 'org.grails.grails-publish'
--- a/src/e2eTest/resources/publish-projects/other-artifacts/groovy-doc-disabled/build.gradle
+++ b/src/e2eTest/resources/publish-projects/other-artifacts/groovy-doc-disabled/build.gradle
@@ -21,7 +21,7 @@ apply plugin: 'java-library'
 apply plugin: 'groovy'
 
 dependencies {
-    implementation 'org.apache.groovy:groovy-all:4.0.24'
+    implementation "org.apache.groovy:groovy-all:$groovyVersion"
 }
 
 groovydoc.enabled = false
--- a/src/e2eTest/resources/publish-projects/other-artifacts/groovy-doc-disabled/build.gradle
+++ b/src/e2eTest/resources/publish-projects/other-artifacts/groovy-doc-disabled/build.gradle
@@ -1,0 +1,40 @@
+buildscript {
+    repositories {
+        maven { url "${System.getenv('LOCAL_MAVEN_PATH')}" }
+        maven { url = 'https://repo.grails.org/grails/core' }
+    }
+    dependencies {
+        classpath "org.grails:grails-gradle-plugin:$grailsGradlePluginVersion"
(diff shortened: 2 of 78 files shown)


src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -218,16 +222,44 @@ Note: if project properties are used, the properties must be defined prior to ap
...
+                it.withSourcesJar()
+            }
+            //Task groovyDocTask = taskContainer.findByName('groovydoc')
+            taskContainer.named('javadocJar', Jar).configure { Jar task ->
+                project.logger.lifecycle("Configuring javadocJar task for project ${project.name}")
+                Task groovyDocTask = taskContainer.findByName('groovydoc')
+                if(groovyDocTask) {
+                    task.dependsOn groovyDocTask
+
+                    task.duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
jdaugherty, line 239: I wish I could have "replaced" what's in the default javadocJar task with the gorovydoc, but I couldn't figure out how to do that.  This does mean the javadoc gets appied, and then the groovydoc gets applied on top of it.  Ideally, we would just include the groovydoc since it suffices for both.  

I think this solution is still better than before.  Both documentation sets will technically be in that jar file and there's a significant amount of test coverage to ensure class specific coverage is present if someone figures out how to fix this in the longer term.

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
```diff
@@ -218,16 +222,44 @@ Note: if project properties are used, the properties must be defined prior to ap
...
+                    task.from groovyDocTask.outputs
+                }
+            }
+
+            taskContainer.named('sourcesJar', Jar).configure { Jar task ->
+                SourceSetContainer sourceSets = SourceSets.findSourceSets(project)
+                task.duplicatesStrategy = DuplicatesStrategy.INCLUDE
+                // don't only include main, but any source set
+                task.from sourceSets.collect { it.allSource }
```
jdaugherty, line 248: I spoke to @matrei and this ensures that all source sets for all source types gets included.  This does mean the resources end up in the jar too, but that's intentional.

src/e2eTest/resources/publish-projects/other-artifacts/explicit-jar-creation-without-gradle-assistance/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/groovy-doc-disabled/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/groovy-only-project/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/java-already-configured/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/multi-project-plugins-applied-child/subproject1/build.gradle
```diff
@@ -0,0 +1,5 @@
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/multi-project-plugins-applied-child/subproject2/build.gradle
```diff
@@ -0,0 +1,5 @@
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/multi-project-plugins-applied-parent/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+
+    apply plugin: 'java-library'
+    apply plugin: 'groovy'
+
+    repositories {
+        maven { url = 'https://repo.grails.org/grails/core' }
+    }
+
+    dependencies {
+        implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/non-groovy-java-sources-included/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

src/e2eTest/resources/publish-projects/other-artifacts/simple-project/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+
+version "${projectVersion}"
+group "org.grails.example"
+
+apply plugin: 'java-library'
+apply plugin: 'groovy'
+
+dependencies {
+    implementation 'org.apache.groovy:groovy-all:4.0.24'
```
matrei: Is it possible to not hard code the groovy version?

matrei (approved):
Could we resolve the groovy version from somewhere or at least DRY it up with a variable?

src/main/groovy/org/grails/gradle/plugin/publishing/GrailsPublishGradlePlugin.groovy
jdaugherty: @jamesfredley if you can figure out another solution to this problem, changes are welcome. In the mean time, this at least gets both versions of the source there and we can improve on this longer term.

jamesfredley:
https://github.com/gradle/gradle/issues/32558

PR #403, opened 2025-02-21
Change and standardize profile task names so it can be used w/ Gradle Java Plugin

### LINKED ISSUE #402: Profile Publishing broken
```console
Grails Publish Plugin requires the Java Plugin to be applied to the project.
```

If you apply the Java Plugin:
```console
A problem occurred evaluating root project 'grails-profiles'.
> Failed to apply plugin 'org.grails.grails-profile'.
   > Cannot add task 'processResources' as a task with that name already exists.
```

jamesfredley:
Allows Java Gradle plugin to be applied via Groovy Gradle plugin for https://github.com/grails/grails-profiles/pull/8

Prior to this change the Java plugin could not be applied to Profile projects.


jdaugherty:
This fixes it so it works with the java/groovy plugins, but those shouldn't be required.  https://github.com/grails/grails-gradle-plugin/pull/404 will make it so it's not required and they can work side by side with both of these changes.

jdaugherty (approved):

PR #406, opened 2025-02-24
Add dependsOn Jar tasks for Profile publishing tasks

jamesfredley:
Resolves the following during Profile publishing:

```
> Task :base:jar FAILED


FAILURE: Build failed with an exception.
[Incubating] Problems report is available at: file:///home/runner/work/grails-profiles/grails-profiles/build/reports/problems/problems-report.html

* What went wrong:
A problem was found with the configuration of task ':base:jar' (type 'Jar').
  - Gradle detected a problem with the following location: '/home/runner/work/grails-profiles/grails-profiles/base/build/libs/base-10.0.2-SNAPSHOT.jar'.
    
    Reason: Task ':base:publishMavenPublicationToMavenRepository' uses this output of task ':base:jar' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':base:jar' as an input of ':base:publishMavenPublicationToMavenRepository'.
      2. Declare an explicit dependency on ':base:jar' from ':base:publishMavenPublicationToMavenRepository' using Task#dependsOn.
      3. Declare an explicit dependency on ':base:jar' from ':base:publishMavenPublicationToMavenRepository' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/8.12/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
```

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
@@ -25,8 +25,6 @@ import org.gradle.api.artifacts.SelfResolvingDependency
 import org.gradle.api.publish.maven.MavenPom
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.publish.maven.tasks.GenerateMavenPom
-import org.gradle.api.publish.maven.tasks.PublishToMavenLocal
-import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
 import org.gradle.api.tasks.bundling.Jar
 import org.grails.gradle.plugin.publishing.GrailsPublishGradlePlugin
 
--- a/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
@@ -24,6 +24,7 @@ import org.gradle.api.artifacts.DependencySet
 import org.gradle.api.artifacts.SelfResolvingDependency
 import org.gradle.api.publish.maven.MavenPom
 import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.publish.maven.tasks.GenerateMavenPom
 import org.gradle.api.publish.maven.tasks.PublishToMavenLocal
 import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
 import org.gradle.api.tasks.bundling.Jar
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
```diff
@@ -52,6 +54,16 @@ class GrailsProfilePublishGradlePlugin extends GrailsPublishGradlePlugin {
             jar.setDescription('Assembles a jar archive containing the profile javadoc.')
             jar.setGroup(BUILD_GROUP)
         })
+
+        project.afterEvaluate { evaluated ->
+            evaluated.tasks.withType(PublishToMavenLocal).each { publishToMavenLocalTask ->
+                publishToMavenLocalTask.dependsOn(project.tasks.withType(Jar))
```
jdaugherty: I see why this change works, but I think it may have unexpected consequences - all jar tasks will run instead of only the ones that are needed.  Given that the profile plugin is still written to look at files instead of gradle dependencies, this is probably our best option.

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
```diff
@@ -52,6 +54,16 @@ class GrailsProfilePublishGradlePlugin extends GrailsPublishGradlePlugin {
             jar.setDescription('Assembles a jar archive containing the profile javadoc.')
             jar.setGroup(BUILD_GROUP)
         })
+
+        project.afterEvaluate { evaluated ->
+            evaluated.tasks.withType(PublishToMavenLocal).each { publishToMavenLocalTask ->
```
jdaugherty: You can collapse these dependencies into one by having the pom generation depend on the jar task instead of the end publish task.

jdaugherty (approved):

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
jamesfredley: Ok, when there is time for a cleanup path, this can be adjusted.

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfilePublishGradlePlugin.groovy
jamesfredley: Updated.  Thank you, I looked for a common Type, which is AbstractPublishToMaven, but you cannot use tasks.withType() with an Abstract class.  I didn't think to set dependsOn with something they all depended on and overlooked GenerateMavenPom tasks, partly because they were working.

jdaugherty (approved):

jdaugherty (approved):

PR #409, opened 2025-02-26
Make profileJar task dependent on java plugin jar task 

### LINKED ISSUE #408: Grails-Shell CLI: One or more required features were not found on the classpath
Occurs when running `grails create-app Test41` directly and via interactive mode `grails` and then `create-app Test41`

```
grails> exit
C:\Users\james\Downloads\7test\bin\Test42>C:\Users\james\Downloads\7test\bin\grails create-app Test42 --stacktrace --ver
bose
| Warning Unrecognized flag: stacktrace.
| Warning Unrecognized flag: verbose.
| Error Error occurred running Grails CLI: One or more required features were not found on the classpath. Required features: [gsp, asset-pipeline]
java.lang.IllegalStateException: One or more required features were not found on the classpath. Required features: [gsp, asset-pipeline]
        at org.grails.cli.profile.AbstractProfile.getRequiredFeatures(AbstractProfile.groovy:244)
        at org.grails.cli.profile.commands.CreateAppCommand.evaluateFeatures(CreateAppCommand.groovy:526)
        at org.grails.cli.profile.commands.CreateAppCommand.handle(CreateAppCommand.groovy:238)
        at org.grails.cli.profile.commands.CreateAppCommand.handle(CreateAppCommand.groovy:385)
        at org.grails.cli.GrailsCli.executeCommandWithArgumentValidation(GrailsCli.groovy:287)
        at org.grails.cli.GrailsCli.execute(GrailsCli.groovy:263)
        at org.grails.cli.GrailsCli.main(GrailsCli.groovy:160)
| Error Error occurred running Grails CLI: One or more required features were not found on the classpath. Required features: [gsp, asset-pipeline]
```

If you run in interactive mode, you then run into https://github.com/grails/grails-profiles/issues/5

https://github.com/grails/grails-core/blob/20793bfe785c21936d519965e6893b6fc2f508b4/grails-shell/src/main/groovy/org/grails/cli/profile/AbstractProfile.groovy#L241-L247

jamesfredley:
to resolve issue with base profile missing the resources from build.

### DIFF
--- a/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
@@ -108,7 +108,7 @@ class GrailsProfileGradlePlugin implements Plugin<Project> {
         })
 
         def profileJarTask = project.tasks.register("profileJar", Jar, { Jar jar ->
-            jar.dependsOn(processProfileResourcesTask, compileProfileTask)
+            jar.dependsOn(processProfileResourcesTask, compileProfileTask, project.tasks.findByName("jar"))
             jar.from(resourcesDir)
             jar.from(classsesDir)
             jar.destinationDirectory.set(new File(project.layout.buildDirectory.getAsFile().get(), "libs"))
--- a/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
+++ b/src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
@@ -108,7 +108,13 @@ class GrailsProfileGradlePlugin implements Plugin<Project> {
         })
 
         def profileJarTask = project.tasks.register("profileJar", Jar, { Jar jar ->
-            jar.dependsOn(processProfileResourcesTask, compileProfileTask, project.tasks.findByName("jar"))
+            jar.dependsOn(processProfileResourcesTask, compileProfileTask)
+
+            def jarTask = project.tasks.findByName("jar")
+            if(jarTask){
+                jar.dependsOn(jarTask)
+            }
+
             jar.from(resourcesDir)
             jar.from(classsesDir)
             jar.destinationDirectory.set(new File(project.layout.buildDirectory.getAsFile().get(), "libs"))


src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
```diff
@@ -108,7 +108,7 @@ class GrailsProfileGradlePlugin implements Plugin<Project> {
         })
 
         def profileJarTask = project.tasks.register("profileJar", Jar, { Jar jar ->
-            jar.dependsOn(processProfileResourcesTask, compileProfileTask)
+            jar.dependsOn(processProfileResourcesTask, compileProfileTask, project.tasks.findByName("jar"))
```
jdaugherty: Will this error if the task doesn't exist?  I assume you only want to add it if it's not null.

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
```diff
@@ -108,7 +108,7 @@ class GrailsProfileGradlePlugin implements Plugin<Project> {
         })
 
         def profileJarTask = project.tasks.register("profileJar", Jar, { Jar jar ->
```
jdaugherty, line 110: We probably should register these tasks in afterEvaluate.  Otherwise if the jar task is registered after the apply of this plugin, it won't be correct.

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
jamesfredley: It will error if the task does not exist, updating for profiles that do not use the java gradle plugin.

src/main/groovy/org/grails/gradle/plugin/profiles/GrailsProfileGradlePlugin.groovy
jamesfredley: I went down this path and it results in having to use afterEvaluate{} in the profile build anywhere that requires dependsOn() from parent profile's task, to resolve errors like the following.  Since this complicates child profile's build, I am leaning towards not making this change.

```
A problem occurred evaluating root project 'grails-profiles'.
> Could not get unknown property 'compileProfile' for project ':angular' of type org.gradle.api.Project.
```

jdaugherty: We should just leave for now then.

jdaugherty (approved):

